PR #1, opened 2015-11-30
[#109031602] bosh-init container

saliceti:
# What
Story: https://www.pivotaltracker.com/story/show/109031602

Create a simple container which contains the latest stable version of bosh-init (0.0.80), plus any dependencies such as Ruby for the AWS CPI, and could be run from Concourse.

All external dependencies are supposed to be mounted in one or several volumes.

# How to review
* Install Docker on the bastion host or any host in the VPC following: https://docs.docker.com/engine/installation/ubuntulinux/
* Tested with Docker 1.9.1
* Follow the instructions in the README to build MicroBOSH

# Who can review
Anyone but @saliceti or @actionjack


timmow:
Testing the build of the container on my locally running boot2docker seems to work fine - i've not tested past seeing the bosh init command is available.

I think it might be good to be consistent with [the terraform image](https://github.com/alphagov/paas-docker-terraform#how-to-run-and-build) and specify how to build in the Readme

Its not immediately clear that this repo will potentially contain multiple Dockerfiles in multiple directories - perhaps we could add some clarification in the main repo Readme?

Could these commits be squashed into 1 commit? There doesn't seem much value in the separate commits.

saliceti:
I have addressed the issues in both READMEs.

I am not sure about squashing the commits, because each one touches a different thing with a clear purpose.

PR #6, opened 2015-12-03
[#109031802] spruce container

saliceti:
# What
Build a container with spruce to be able to merge manifests on Concourse.

It uses a go alpine linux image as base to be able to compile go code while keeping a small image size.

# How to review
### Requirements: 
* Run docker locally
* Set docker environment variables if necessary. Example with docker-machine run:

```
eval "$(docker-machine env <machine>)"`
```

### Run

* `bundle install`
* `rake build:spruce`
* `rake spec:spruce` will run the spruce test
* `rake` will run all the tests
* check the README for usage instructions

### Who can review
Anyone but @saliceti or @jimconner

### DIFF
--- a/spruce/Dockerfile
+++ b/spruce/Dockerfile
@@ -1,0 +1,6 @@
+FROM golang:1.5.1-alpine
+
+RUN apk update
+RUN apk add git \
+  && go get github.com/geofffranks/spruce \
+  && apk del git
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ services:
   - docker
 
 before_install:
-  - cd bosh-init && docker build -t bosh-init .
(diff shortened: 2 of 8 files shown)


spruce/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.5.1-alpine
+
+RUN apk update
+RUN apk add git
+RUN go get github.com/geofffranks/spruce
+RUN apk del git
```
actionjack: You may want to switch the last 3 commands to just 
``` RUN apk add git && go get github.com/geofffranks/spruce && apk del git```` this will produce a single layer that will produce a smaller image

actionjack:
Nice PR! :+1:

PR #7, opened 2016-01-05
[#109060102] Add new container curl-ssl

keymon:
[#109060102 Add new container curl-ssl](https://www.pivotaltracker.com/story/show/109060102)

# What

Implement a new container curl-ssl which includes the curl command with SSL support and openssl binary.

The container is based on Alpine:3.3

The spec check won't use the serverspec package resource because it does
not support alpine. Instead, we use the `apk` command directly.

# Dependencies

This PR depends on #8 being merged first, to pin the spruce version (merged already)

# How to test it

Using the rake file (I recommend use of [RVM](http://getrvm.io)):

```
bundle install
bundle exec rake build:curl_ssl spec:curl_ssl
```

You can also test the commands:

```
docker run -i -t curl-ssl curl https://www.google.com
docker run -i -t curl-ssl openssl
```

# Who?

Anyone but @keymon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install:
   - bundle install
   - bundle exec rake build:bosh_init
   - bundle exec rake build:spruce
+  - bundle exec rake build:curl_ssl
 
 script:
   - cd ${TRAVIS_BUILD_DIR} && bundle exec rake
--- a/README.md
+++ b/README.md
@@ -8,3 +8,4 @@ List:
 
 * [bosh-init](https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/bosh-init)
 * [spruce](https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/spruce)
+* [curl-ssl](https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/master/curl-ssl)
--- a/Rakefile
+++ b/Rakefile
@@ -10,11 +10,15 @@ namespace :build do
     system "cd spruce && docker build -t spruce ."
   end
 
+  desc "Build curl-ssl image"
+  task :curl_ssl do
+    system "cd curl-ssl && docker build -t curl-ssl ."
+  end
 end
 
 require 'rspec/core/rake_task'
 
-task :spec => ["spec:bosh-init", "spec:spruce"]
+task :spec => ["spec:bosh-init", "spec:spruce", "spec:curl_ssl"]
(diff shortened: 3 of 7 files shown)


keymon:
In this branch I implement install spruce from the release to pin the version: https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/spruce_from_release but until [this issue](https://github.com/geofffranks/spruce/issues/83) is not fixed, we can keep the compiled version.

dcarley:
> In this branch I implement install spruce from the release to pin the version: https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/spruce_from_release but until this issue is not fixed, we can keep the compiled version.

I think we should pin it with the following:
```
diff --git a/spec/spruce/spruce_spec.rb b/spec/spruce/spruce_spec.rb
index c7a6970..bd2e9a7 100644
--- a/spec/spruce/spruce_spec.rb
+++ b/spec/spruce/spruce_spec.rb
@@ -29,7 +29,7 @@ describe "spruce image" do
   end

   it "has the spruce version #{SPRUCE_VERSION}" do
-    expect(spruce_version).to eq("spruce - Version #{SPRUCE_VERSION} (master)")
+    expect(spruce_version).to eq("spruce - Version #{SPRUCE_VERSION}")
   end

   def spruce_version
diff --git a/spruce/Dockerfile b/spruce/Dockerfile
index 7d8d858..a5e3a8d 100644
--- a/spruce/Dockerfile
+++ b/spruce/Dockerfile
@@ -1,6 +1,11 @@
 FROM golang:1.5.1-alpine

+ENV SPRUCE_VERSION v0.12.0
+
 RUN apk update
 RUN apk add git \
   && go get github.com/geofffranks/spruce \
+  && cd ${GOPATH}/src/github.com/geofffranks/spruce \
+  && git checkout ${SPRUCE_VERSION} \
+  && go install \
   && apk del git
```

The change to the test is required and indicates that we're not actually building the released version - it's the last release + whatever has landed on the master branch since. It won't pin dependencies, so we should still move to a binary in the near future.

keymon:
@dcarley +1 I will create a different PR

keymon:
Added #8 which should be merged first

keymon:
#8 merged, and this PR has been rebased.

dcarley:
Martin noted in Slack - this needs a rebase again.

keymon:
Rebased again after #9

actionjack:
:+1: LGTM

PR #10, opened 2016-01-05
Improve error message when tests can't find image

dcarley:
It momentarily confused me why the tests wouldn't work locally. I've removed
the `should be available` test because this doesn't report the correct thing
when `@image` is `nil`. Because of this we no longer need `@image`, we just
want the ID.

Before:

    1) bosh-init image should be available
       Failure/Error: set :docker_image, @image.id
       NoMethodError:
         undefined method `id' for nil:NilClass

       # ./spec/bosh-init/bosh-init_spec.rb:8:in `block (2 levels) in <top (required)>'

After:

    1) bosh-init image installs the right version of Ubuntu
       Failure/Error: raise "Docker image '#{name}' not found. You may need to run 'rake build:<name>' first" unless image
       RuntimeError:
         Docker image 'bosh-init:latest' not found. You may need to run 'rake build:<name>' first

       # ./spec/spec_helper.rb:20:in `find_image_id'
       # ./spec/bosh-init/bosh-init_spec.rb:7:in `block (2 levels) in <top (required)>'

### DIFF
--- a/spec/bosh-init/bosh-init_spec.rb
+++ b/spec/bosh-init/bosh-init_spec.rb
@@ -4,13 +4,8 @@ require 'serverspec'
 
 describe "bosh-init image" do
   before(:all) {
-    @image = Docker::Image.all().detect{|i| i.info['RepoTags'].include? 'bosh-init:latest' }
-    set :docker_image, @image.id
+    set :docker_image, find_image_id('bosh-init:latest')
   }
-
-  it "should be available" do
-    expect(@image).to_not be_nil
-  end
 
   it "installs the right version of Ubuntu" do
     expect(os_version).to include("Ubuntu 14")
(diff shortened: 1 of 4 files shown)


mtekel:
What is still confusing is that we call it `bosh-init` everywhere, but the rake task is `bosh_init` - due to ruby hash not liking hyphens.

mtekel:
This doesn't work when I get image from docker (`docker pull governmentpaas/spruce`) instead of building locally, as the image is then called `governmentpaas/spruce`.

dcarley:
> This doesn't work when I get image from docker (docker pull governmentpaas/spruce) instead of building locally, as the image is then called governmentpaas/spruce.

I'm not sure that it's meant to or whether it is advantageous to do so.

mtekel:
I suppose you don't always need/want to build all images locally if you need to do anything. Concourse doesn't - it pulls them from docker. If I want to test with exactly same containers locally, I also pull them from docker. Yet then the tests fail.

actionjack:
LGTM:

```bash
 Failure/Error: raise "Docker image '#{name}' not found. You may need to run 'rake build:<name>' first" unless image
     RuntimeError:
       Docker image 'curl-ssl:latest' not found. You may need to run 'rake build:<name>' first
```

PR #12, opened 2016-01-13
[#111137450] Cache bosh-init AWS CPI compilation in image

alext:
## Problem

Every time bosh-init runs in concourse it has to compile the CPI, which takes 3-5 minutes

## Solution

This adds a step to the build for the bosh-init container to run a dummy deploy which in turn causes the CPI to be compiled and cached in `/root/.bosh_init`.

## Issues/limitations

The cached CPI version has to match the version being used in the manifests used by bosh-init for the cache to be used.

Secondly, the compilation cache is keyed from the `installation_id` which is written to the bosh state.json file. bosh-init generates a random id for new deployments that don't yet have a state file. To work around this, this creates a state.json file containing a predefined installation_id before running bosh-init.  To use the cached version it's then necessary for all runs of bosh-init to use this same `installation_id` (see alphagov/paas-cf#37).

### DIFF
--- a/bosh-init/README.md
+++ b/bosh-init/README.md
@@ -52,3 +52,12 @@ $ docker run -v /home/ubuntu/docker-bosh-init-manifest:/docker-bosh-init-manifes
 bosh-init deploy /docker-bosh-init-manifest/bosh-manifest.yml
 
 ```
+
+## Caching compile of AWS CPI
+
+The container has been preseeded with a compile of the bosh AWS CPI. If your
+manifest is using the same version, this will speed things up considerably.
+
+The version of the CPi that's cached is controller by the `BOSH_AWS_CPI_URL`
+and `BOSH_AWS_CPI_CHECKSUM` variables in the `Dockerfile`, which should contain
+values that match the `url` and `sha1` parameters of the bosh release.
--- a/bosh-init/bosh_init_cache/minimal-state.json
+++ b/bosh-init/bosh_init_cache/minimal-state.json
@@ -1,0 +1,3 @@
+{
+  "installation_id": "44f01911-a47a-4a24-6ca3-a3109b33f058"
+}
--- a/bosh-init/bosh_init_cache/minimal_in.yml
+++ b/bosh-init/bosh_init_cache/minimal_in.yml
@@ -1,0 +1,24 @@
+---
+name: dummy-deployment
+cloud_provider:
+  template:
+    name: aws_cpi
+    release: bosh-aws-cpi
+releases:
+- name: bosh-aws-cpi
+  sha1: BOSH_AWS_CPI_CHECKSUM
+  url: BOSH_AWS_CPI_URL
+networks:
+- name: dummy
+  type: vip
+resource_pools:
+- name: dummy
+  network: dummy
+  stemcell:
+    sha1: 1a29c43d4e8abf7476ed6bb83168df1bdb742022
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3074
+jobs:
+- name: dummy
+  networks:
+  - name: dummy
+  resource_pool: dummy
--- a/spec/bosh-init/bosh-init_spec.rb
+++ b/spec/bosh-init/bosh-init_spec.rb
@@ -34,11 +34,18 @@ describe "bosh-init image" do
   end
 
   it "contains the compiled CPI packages" do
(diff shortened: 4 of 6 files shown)


dcarley:
I think this is *probably* worth doing. At least until `bosh-init` supports using S3 as a package cache in cloudfoundry/bosh-init#19.

I'm going to comment on some individual lines. I'm not assigning it to myself though - would like more comments from other people in the team.

bosh-init/README.md
```diff
@@ -52,3 +52,12 @@ $ docker run -v /home/ubuntu/docker-bosh-init-manifest:/docker-bosh-init-manifes
...
 ```
+
+## Caching compile of AWS CPI
+
+The container has been preseeded with a compile of the bosh AWS CPI. If your
+manifest is using the same version, this will speed things up considerably.
+
+The version of the CPi that's cached is controller by the `BOSH_AWS_CPI_URL`
+and `BOSH_AWS_CPI_CHECKSUM` variables in the `Dockerfile`, which should contain
+vales that match the `url` and `sha1` parameters of the bosh release.
```
dcarley: Typo `s/vales/values/`

bosh-init/bosh_init_cache/minimal-state.json
```diff
@@ -0,0 +1,3 @@
+{
+  "installation_id": "44f01911-a47a-4a24-6ca3-a3109b33f058"
```
dcarley: Can you add to the commit message how this was generated? Is it significant?

I thought it might be a [V4 (random) UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29) but it doesn't seem to be.

bosh-init/bosh_init_cache/minimal_in.yml
```diff
@@ -0,0 +1,33 @@
...
+  template:
+    name: aws_cpi
+    release: bosh-aws-cpi
+releases:
+- name: bosh-aws-cpi
+  sha1: BOSH_AWS_CPI_CHECKSUM
+  url: BOSH_AWS_CPI_URL
+networks:
+- name: dummy
+  type: manual
```
dcarley: This could be simplified to:
```
- name: dummy
  type: vip
```

bosh-init/bosh_init_cache/minimal_in.yml
```diff
@@ -0,0 +1,33 @@
...
+    gateway: 192.168.0.1
+resource_pools:
+- name: dummy
+  network: dummy
+  stemcell:
+    sha1: 1a29c43d4e8abf7476ed6bb83168df1bdb742022
+    url: https://bosh.io/d/stemcells/bosh-aws-xen-hvm-ubuntu-trusty-go_agent?v=3074
+jobs:
+- name: dummy
+  instances: 0
```
dcarley: I think this can be simplified to:
```
jobs:a
- name: dummy
  networks:
  - name: dummy
  resource_pool: dummy
```

spec/bosh-init/bosh-init_spec.rb
```diff
@@ -33,6 +33,14 @@ def os_version
     expect(bosh_init_version).to eq("version #{BOSH_INIT_VERSION}")
   end
 
+  it "contains the compiled CPI packages" do
+    cmd = command('cat /root/.bosh_init/installations/*/compiled_packages.json')
+    expect(cmd.exit_status).to eq(0)
+    compiled_packages = JSON.parse(cmd.stdout)
+    cpi_package = compiled_packages.find {|p| p["Key"]["PackageName"] == "bosh_aws_cpi" }
```
dcarley: Should we check that the blobs exist too?

bosh-init/bosh_init_cache/minimal-state.json
alext: This was just an arbitrary value taken from one both-init run. I've updated the commit message accordingly.

bosh-init/README.md
alext: Fixed.

mtekel:
The fact that it has to be done this way is just because how bosh-init works. If it had different/better behaviour, the PRs could be nicer. Nevertheless this fixes the problem in the most nicest and clean way. For me this is a +1 and should be merged.

alext:
I've updated this to reduce the minimal manifest further, and also extended the tests a bit.

dcarley:
Only other feedback from Slack:
> @dcarley: I think it is a bunch of hacks,, but saves precious time :simple_smile:

OK, let's do this. I'll test it out.

PR #13, opened 2016-01-19
[#110059806] Secrets generation

saliceti:
# What
Container with utilities to generate secrets like passwords and possibly SSL and SSH keys in the future. This is required for generate-secrets tasks in concourse pipelines.

# How to review
* Configure a local environment with docker and [serverspec](http://serverspec.org/)
* Build with `rake build:mksecrets`
* Test with `rake spec:mksecrets`
* Run commands like `docker run -i -t mksecrets pwgen` and
`docker run -i -t mksecrets sh -c "echo mysecret | mkpasswd -m sha-512"`

# Who can review
Anyone but me

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
   - bundle exec rake build:bosh_init
   - bundle exec rake build:spruce
   - bundle exec rake build:curl_ssl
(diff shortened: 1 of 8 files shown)


keymon:
Cool, LGTM. Merging

PR #15, opened 2016-02-02
[#112819029] Implement automated task generation in Rakefile

keymon:
https://www.pivotaltracker.com/story/show/112819029

# What

Currently add containers to this repo is tedious and there is lots of repetition. We want to simplify it.

# Context

We refactor the repository to:
 * Automatically generate all the tasks to build and test the containers in the Rakefile based on existing Dockerfiles and spec files.
 * Call the new tasks from travis
 * Refactor the tests to keep all the constants and logic in the same file
 * Update the README to explain how to add containers and remove unnecessary duplication.

# How to test this

Travis builds and runs the tests of the containers, and it should pass for all the containers.

You can also run it manually (docker required):

```
rvm use --create 2.2.2@containers
bundle install
bundle exec rake
```

# who?

Anyone but @keymon


Rakefile
```diff
@@ -1,61 +1,35 @@
...
-  end
-
-  desc "Run all specs for awscli"
-  RSpec::Core::RakeTask.new("awscli") do |t|
-    t.pattern = "spec/awscli/**/*_spec.rb"
+namespace :spec do |ns|
+  require 'rspec/core/rake_task'
+
+  containers.each { |container|
+    spec_pattern = File.join(container, "**", "*spec.rb")
```
alext: Should this glob be `<container>/**/*_spec.rb`  It's convention with Rspec that all specs are in files named `foo_spec.rb` etc. It would be good to stick with that convention here.

keymon: @alext I agree, updated.

combor:
Not sure if this should be a part of this PR but I would like to have short guide how to add new container and tests to this repo.

alext:
Tested, and this all works correctly.

keymon:
@combor, @alext repo updated with your suggestions

dcarley:
:+1:

PR #16, opened 2016-02-04
[#111428758] Add a container certstrap CA manager tool

keymon:
[Auto-generate SSLs for all internal components](https://www.pivotaltracker.com/story/show/111428758)

What
----

We want to generate the certificates for all the components, signed by a CA which is trusted by all the related VMs.

For that, we will use [certstrap](https://github.com/square/certstrap/) tooling to simplify management of the certificates.
In the concourse job pipelines we will also interact with S3 using the script `s3init.sh`, so we must include `curl` and `ca-certificates`.

how to test:
-----------

Travis tests should execute the spec added in this PR and pass .
```
rake build:certstrap spec:certstrap
```

Notes
-----

A docker hub automated build must be setup with this Dockerfile.

The PR https://github.com/alphagov/paas-cf/pull/92 must be updated to use
the new container

Who?
----

Anyone but @keymon or @mtekel

### DIFF
--- a/certstrap/Dockerfile
+++ b/certstrap/Dockerfile
@@ -3,7 +3,7 @@ FROM golang:1.5.3-alpine
 ENV GIT_COMMIT 8b3cd66cfe2433e2e998c46a618eb09f732c2c2c
 ENV CERTSTRAP_VERSION 0.1.0
 
-RUN apk add --update git bash openssl \
+RUN apk add --update git bash openssl curl ca-certificates \
   && go get -d github.com/square/certstrap \
   && cd ${GOPATH}/src/github.com/square/certstrap \
   && git checkout ${GIT_COMMIT} \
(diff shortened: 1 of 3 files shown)


certstrap/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV GIT_COMMIT 8b3cd66cfe2433e2e998c46a618eb09f732c2c2c
+ENV CERTSTRAP_VERSION 0.1.0
+
+RUN apk add --update git bash openssl curl ca-certificates \
+  && go get -d github.com/square/certstrap \
+  && cd ${GOPATH}/src/github.com/square/certstrap \
+  && git checkout ${GIT_COMMIT} \
+  && ./build \
+  && cp bin/certstrap /usr/local/bin \
+  && apk del git bash \
```
saliceti: You can add `&& rm -rf ${GOPATH} \`
(this will save 6MB...)

keymon: Don't be cheap!  :smile: 

I think we can change all these containers when we have our own CI system by a pipeline which compiles the binary statically and adds it to a naked alpine container.

saliceti:
Tested successfully

certstrap/Dockerfile
keymon: Done :kiss::dash:

PR #17, opened 2016-02-08
[#112503901] Create cf-cli image

alext:
This contains the CF CLI, and a Go runtime. This will be used to run the
acceptance tests in concourse.


mtekel:
works fine, passes tests

PR #18, opened 2016-02-09
[#112503901] Test for curl in cf-cli container

alext:
The [CF acceptance-test helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers) depend on curl, so we should verify it's presence in the container.


alext:
Going to rename this container...

PR #23, opened 2016-02-16
[#112908839] Update bosh-cli openssl version.

alext:
httpclient (which is a dependency of bosh-cli) has a [check for the
openssl version](https://github.com/nahi/httpclient/blob/v2.7.1/lib/httpclient/ssl_config.rb#L441-L452) to ensure it includes fixes for a vulnerability.
Versions below this cause it to emit warnings.

Because debian backports security fixes instead of upgrading openssl,
these warnings are triggered here, even though openssl is no longer
vulnerable.

Digging further, the CVE mentioned in the comment in httpclient
specifies that the only versions affected are 1.0.1n, 1.0.1o, 1.0.2b and
1.0.2c, so the httpclient check identifies non-vulnerable versions as
vulnerable.

To work around this, we've switched the bosh-cli container to be based
on alpine, which includes the newer versions of openssl.

### DIFF
--- a/bosh-cli/Dockerfile
+++ b/bosh-cli/Dockerfile
@@ -1,5 +1,5 @@
-FROM ruby:2.2-slim
+FROM ruby:2.2-alpine
 
 RUN gem install bosh_cli -v 1.3192.0 --no-rdoc --no-ri
-RUN apt-get update && apt-get install -y openssh-client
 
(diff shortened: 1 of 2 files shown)


dcarley:
Tested `deploy-cloudfoundry` and `destroy-cloudfoundry` with alphagov/paas-cf@c71c00c19850d22e035cf0a4976f070d369ff4eb

PR #24, opened 2016-02-17
[#110652158] Add `file` utility to bosh-cli container.

alext:
Without this, we are seeing warnings from `bosh upload relase` saying
"sh: file: not found". Despite this, the bosh upload seems to still
work.

Given this is evidently a dependency of the bosh-cli it seems better
that we include it.

Came across this while looking at making the release upload process
deal with more than one version of a release.


combor:
Looks good and Travis build passes. Merging...

PR #25, opened 2016-02-23
[#114193485] Add terraform container from other old repository and enable postgresql

keymon:
What
----

We want to be able to create roles and DBs in postgres using terraform to create
the DBs for the RDS based CF databases.

As we add this change, we will import the separated repository
in https://github.com/alphagov/paas-docker-terraform to
this repository, as this is where we are implementing all the containers.

Context
-------

 * I want to import the code and history from https://github.com/alphagov/paas-docker-terraform
 * and include the postgres provider

How to test?
------------

As usual in this repo, by travis or:

```
rvm use --create 2.2.2@containers
bundle install
bundle exec rake
```

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/terraform_spec.rb
+++ b/terraform/terraform_spec.rb
@@ -1,0 +1,45 @@
+require 'spec_helper'
+require 'docker'
+require 'serverspec'
+
+describe "Terraform image" do
+  before(:all) {
+    set :docker_image, find_image_id('terraform:latest')
+  }
+
+  it "installs the Alpine linux" do
+    expect(file("/etc/alpine-release")).to be_file
+  end
+
+  it "installs Root Certificates" do
+    expect(file("/usr/share/ca-certificates/mozilla/GlobalSign_Root_CA.crt")).to be_file
+  end
+
+  it "checks if terraform binary is executable" do
+    expect(file("/usr/local/bin/terraform")).to be_mode 755
+  end
+
+  it "has the Terraform version 0.6.10" do
+    expect(terraform_version).to include("Terraform v0.6.10")
+  end
+
+  def terraform_version
+    command("terraform version").stdout.strip
+  end
+
+  it "installs SSH" do
+    expect(ssh_version).to include("OpenSSH")
+  end
+
+  def ssh_version
+    command("ssh -V").stderr.strip
+  end
+
+  it "should not have binary directory larger than 200M" do
+    expect(binaries_size).to be < 200
+  end
+
+  def binaries_size
+    Integer(command("du -m /usr/local/bin").stdout.split.first)
+  end
+end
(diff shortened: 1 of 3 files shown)


terraform/terraform_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'spec_helper'
+require 'docker'
+require 'serverspec'
+
+describe "Terraform image" do
+    before(:all) {
```
alext: The indentation is off in this file.  The convention for ruby is 2 space indents. I realise that this is how it was previously, but this seems like a good opportunity to fix it.

alext:
A couple of minor points, otherwise this looks good to me. I like how you've kept the history from the original repo.

terraform/terraform_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    it "checks if terraform binary is executable" do
+        expect(file("/usr/local/bin/terraform")).to be_mode 755
+    end
+
+    it "installs all the whitelisted binaries" do
+        command("echo $BINARY_WHITELIST").stdout.split.each { |f|
+            expect(file("/usr/local/bin/#{f}")).to be_mode 755
+        }
```
alext: This test feels a little circular to me. As it stands, nothing will fail if one of the required binaries is removed in the `Dockerfile`.

This is effectively testing that the `Dockerfile` installs what it says it's going to, which is a different thing from testing that it installs the things we need.

keymon: Well, one of the issues I found is that if you do a typo in the Whitelist, it will run clean and don't tell you that it missed one.

But OK, I will add an additional test and also change the title of this one.

keymon:
@alext I did the required changes.

terraform/terraform_spec.rb
dcarley: > Well, one of the issues I found is that if you do a typo in the Whitelist, it will run clean and don't tell you that it missed one.

I was wondering this and tested it this morning. It seems `unzip` will return a non-zero exit code if you specify a path that's not present in the archive. Does that not fail the `docker build`?

keymon: >  It seems unzip will return a non-zero exit code if you specify a path that's not present in the archive. Does that not fail the docker build?

I have just checked, and it fails on my OS (ubuntu, UnZip 6.00 of 20 April 2009, by Debian. Original by Info-ZIP.) but not on the alpine one (BusyBox v1.24.1 (2015-12-16 08:00:02 GMT) multi-call binary.). 

So the test is kinda required.

dcarley: I was a bit indifferent before, but that definitely sways me :+1:

mtekel:
builds, passes tests and looks otherwise OK

PR #26, opened 2016-02-25
[#114193485] Create psql docker container

saliceti:
# What
Create psql Postgres client container so we can run psql commands from Concourse. This is necessary to set up the CF RDS database.
Contains test and README.

# How to review
* build: `rake build:psql`
* test: `rake spec:psql`
* run: `docker run psql psql --version`

# Who can review
Anyone but @mtekel or me


combor:
LGTM. Merging...

PR #27, opened 2016-02-25
[#113824903] Bump terraform to 0.6.12

alext:
There's an issue with earlier versions when handling a security group
with multiple ingress blocks for the same port range. It results in
terraform updating the resource on every run. This is no longer the case
with 0.6.12. See https://github.com/hashicorp/terraform/issues/5301#issuecomment-188746760 for details of this issue.

See also the related paas-cf PR - https://github.com/alphagov/paas-cf/pull/123

## How to review.

This is best reviewed in conjunction with the above paas-cf PR. See notes there for details.

## Who can review

Anyone but mysql.

### DIFF
--- a/terraform/Dockerfile
+++ b/terraform/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine
 ENV PATH $PATH:/usr/local/bin
-ENV TERRAFORM_VER 0.6.10
+ENV TERRAFORM_VER 0.6.12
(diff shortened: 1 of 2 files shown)


mtekel:
This PR as a "bump version" works correctly. However I still have an issue with SG diffs not matching....

mtekel:
..which is to be expected and workaround is implemented in https://github.com/alphagov/paas-cf/pull/123

PR #28, opened 2016-03-01
[#114186797] Add unzip command to cf-acceptance-tests container

keymon:
[#114186797 Run smoke tests and acceptance tests in concourse task rather than errand](https://www.pivotaltracker.com/story/show/114186797)

What
----

We want to run the cf-acceptance-tests on concourse directly, but as they
[call unzip command](https://github.com/cloudfoundry/cf-acceptance-tests/blob/e5e7b4726f61f9e66da43c1daf0151d6b4492be1/v3/package_test.go#L75),
we need to install the package in the container.

How to test this
----------------

As usual, follow README instructions and:

```
rake build:cf-acceptance-tests spec:cf-acceptance-tests
```

Travis tests should be OK.

Who?
----

Anyone but @keymon

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,5 +1,8 @@
 FROM golang:1.5.3-wheezy
 
 RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.15.0' | tar -zx -C /usr/local/bin
+RUN apt-get update \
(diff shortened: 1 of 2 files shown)


alext:
Tests failed due to a network error while downloading terraform. I've kicked off another build.

Interestingly, the build didn't abort when the build step failed, and continued on to run the tests, which then errored because the container didn't exist.

PR #29, opened 2016-03-04
[#114778351] Remove mksecrets container.

alext:
This is no longer needed since we switched to using ruby to generate
secrets in https://github.com/alphagov/paas-cf/pull/130

Once this is merged, I'll remove the dockerhub build for this container.


combor:
Looks good...merging

PR #30, opened 2016-03-08
[#113749465] Add container with cf-cli andd tools for deploy

keymon:
[#113749465 Make Cloud Foundry Metrics Available](https://www.pivotaltracker.com/story/show/113749465)

What?
-----

We need a container to be able to deploy CF apps in a task. We also need some other tools
like curl, unzip and git to be able to download, uncompress or clone/checkout dependencies.

Container with CF client command and some other tools to help deploying
apps in cloudfoundry: curl, git, unzip.

Based on Debian wheezy, as the distributed binary cf-cli depends on the
libc of a Debian like linux distro (alpine won't work).

Dependencies
-------------------

This is required before https://github.com/alphagov/paas-cf/pull/139 is merged.

Test
----

Travis should run the tests and end successfull.

To run tests locally: `rake build:cf-cli spec:cf-cli`

Who?
----

Anyone but @keymon

### DIFF
--- a/cf-cli/cf-acceptance-tests_spec.rb
+++ b/cf-cli/cf-acceptance-tests_spec.rb
@@ -1,0 +1,35 @@
+require 'spec_helper'
+require 'docker'
+require 'serverspec'
+
+CF_CLI_VERSION="6.15.0"
+
+describe "cf-cli image" do
+  before(:all) {
+    set :docker_image, find_image_id('cf-cli:latest')
+  }
+
+  it "has the expected version of the CF CLI" do
+    expect(
+      command("cf --version").stdout
+    ).to match(/cf version #{CF_CLI_VERSION}/)
+  end
+
+  it "has curl available" do
+    expect(
+      command("curl --version").exit_status
+    ).to eq(0)
+  end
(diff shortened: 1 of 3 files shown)


keymon:
Rebased with #32 so that it should fix the build in redis.

cf-cli/cf-acceptance-tests_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'spec_helper'
+require 'docker'
+require 'serverspec'
+
+CF_CLI_VERSION="6.15.0"
+
+describe "cf-cli imagee" do
```
jimconner: You've spelt 'image' wrong here - Please fix.

keymon: @jimconner done

jimconner:
All works as advertised... Merging.

```
cf-cli image
  has the expected version of the CF CLI
  has curl available
  has unzip available
  has git available

Finished in 0.55171 seconds (files took 0.6334 seconds to load)
4 examples, 0 failures
```

PR #31, opened 2016-03-08
[#115258555] Use the latest spruce 1.1.2 from release

keymon:
What?
----

We have been using the version 0.13.0 of spruce, but compiling it
in the Dockerfile because the release binary version did link to
the debian/ubuntu libc. We were not pinning the git commit of the
spruce code.

But the upstream project has some changes that makes fail the build,
probably new dependencies added which require the use of godep.

As since https://github.com/geofffranks/spruce/issues/83 the binary
in the releases is OK and can work in a plain alpine, we can create
a container pulling the binary release. This will simplify our
container, make it smaller and pin the version number.

How to test this?
----------------

Travis should run OK and you can run the tests with:

```
rake build:spruce spec:spruce
```

who?
----

anyone but @keymon or @henrytk

### DIFF
--- a/spruce/Dockerfile
+++ b/spruce/Dockerfile
@@ -1,11 +1,11 @@
-FROM golang:1.5.1-alpine
+FROM alpine:3.3
 
-ENV SPRUCE_VERSION 0.13.0
+ENV SPRUCE_VERSION 1.1.2
 
-RUN apk add --update git \
-  && go get -d github.com/geofffranks/spruce \
-  && cd ${GOPATH}/src/github.com/geofffranks/spruce \
-  && git checkout v${SPRUCE_VERSION} \
-  && go install \
-  && apk del git \
+ENV PACKAGES "curl openssl ca-certificates"
+
+RUN apk add --update $PACKAGES \
+  && curl -f -L https://github.com/geofffranks/spruce/releases/download/v$SPRUCE_VERSION/spruce-linux-amd64 > /usr/local/bin/spruce \
+  && chmod +x /usr/local/bin/spruce \
+  && apk del $PACKAGES \
   && rm -rf /var/cache/apk/*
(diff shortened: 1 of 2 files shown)


spruce/Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-RUN apk add --update git \
-  && go get -d github.com/geofffranks/spruce \
-  && cd ${GOPATH}/src/github.com/geofffranks/spruce \
-  && git checkout v${SPRUCE_VERSION} \
-  && go install \
-  && apk del git \
+ENV PACKAGES "curl openssl ca-certificates"
+
+RUN apk add --update $PACKAGES \
+  && curl -f -L https://github.com/geofffranks/spruce/releases/download/v$SPRUCE_VERSION/spruce-linux-amd64 > /usr/local/bin/spruce \
```
dcarley: Can you brace this variable so that it's easier to read and prevents any interpolation mistakes?
```
https://github.com/geofffranks/spruce/releases/download/v${SPRUCE_VERSION}/spruce-linux-amd64
```

dcarley:
Have you reviewed the CHANELOG for versions 0.13.0...1.1.2 to check any backwards incompatible issues? Have you tested the alphagov/paas-cf tests or pipelines to see whether it still works for our use?

alext:
I'd note that there are releases available for eariler versions as well.  eg 0.13.0 is available here - https://github.com/geofffranks/spruce/releases/tag/v0.13.0 albeit, this is zipped, and would therefore need unzipping.

dcarley:
> I'd note that there are releases available for eariler versions as well. eg 0.13.0 is available here - https://github.com/geofffranks/spruce/releases/tag/v0.13.0 albeit, this is zipped, and would therefore need unzipping.

:+1: 

I'm not against upgrading. But for the purpose of unblocking our build and board, it's probably safer to use the same version.

keymon:
Ok, I will close this one. :)

PR #33, opened 2016-03-11
[#114000923] Add fly-cli container to update pipelines

keymon:
[#114000923 self-updating pipelines](https://www.pivotaltracker.com/story/show/114000923)

What
----

We want to be able to update concourse pipelines from concourse itself, using
the same scripts that we use to setup the pipelines.

Add a fly-cli container to run jobs that update concourse pipelines.
We also require bash, curl, and ruby to support the pipeline scripts.

How to test
-----------

As usual: `make build:fly-cli spec:fly-cli` and travis should success.

Who?
----

Anyone but @keymon or @saliceti

### DIFF
--- a/fly-cli/Dockerfile
+++ b/fly-cli/Dockerfile
@@ -1,0 +1,10 @@
+FROM alpine:3.3
+
+ENV PACKAGES bash ruby curl ca-certificates
+ENV FLY_CLI_URL http://ci.concourse.ci/api/v1/cli?arch=amd64&platform=linux
+
+RUN apk add --update $PACKAGES \
+    && rm -rf /var/cache/apk/* \
+    && curl -f -L "${FLY_CLI_URL}" > /usr/local/bin/fly \
+    && chmod +x /usr/local/bin/fly
+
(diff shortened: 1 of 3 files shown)


dcarley:
Are we going to have trouble with the version of `fly` matching the Concourse server it's talking to? Would it be better to download `fly` within the task (should be fast and will be of the correct version)? Does alphagov/paas-cf@4a7d768391f0505c324e3b17263be58b135d22c3 do that for us anyway?

keymon:
@dcarley actually the client is downloaded anyway in the pipeline config scripts. This one is not used/

But I though it would be nice to keep this here, to have a container with the fly command.

I don't mind delete it.

keymon:
But, actually, my main motivation to download fly cli is to be able to keep naming the container `fly-cli` :angel:

dcarley:
> But, actually, my main motivation to download fly cli is to be able to keep naming the container fly-cli

That's admirable, but I suspect it may be more confusing if the version doesn't matter because it's not actually used.

Is there another container that we can use instead? Do the ruby-slim containers have bash and curl?

keymon:
Sure, I will.

keymon:
Replaced by: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/34

PR #35, opened 2016-03-14
[#114000923] Add self-update-pipelines container to run update pipelines related scripts

keymon:
[#114000923 self-updating pipelines](https://www.pivotaltracker.com/story/show/114000923)

What
----

We want to be able to update concourse pipelines from concourse itself, using
the same scripts that we use to setup the pipelines.

Add a self-update-pipelines container to run jobs which update the pipelines
in concourse using the configuration scripts.
These scripts require bash, curl, ruby and make.

How to test
-----------

As usual: `make build:self-update-pipelines spec:self-update-pipelines` and
travis should success.

Who?
----

Anyone but @keymon

### DIFF
--- a/self-update-pipelines/Dockerfile
+++ b/self-update-pipelines/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.2-slim
+
+ENV PACKAGES make
+
+RUN apt-get update \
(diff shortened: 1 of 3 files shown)


alext:
Created dockerhub build here: https://hub.docker.com/r/governmentpaas/self-update-pipelines/builds/

PR #36, opened 2016-03-14
Make build rake tasks exit non-zero on error

alext:
This will give us early feedback on build errors, and make it more obvious what has actually errored.

At present, if a build fails, the rake task doesn't exit non-zero, so everything else continues, and then the tests fail, and it's not easy to see what the actual problem is.

## How to test

Cause one of the image builds to fail (eg by adding a `RUN false` line to the `Dockerfile`). Then run `bundle exec rake build:all`, and verify that it exits non-zero immediately that it encounters the error.

## Who can test

Anyone but myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,8 @@ sudo: required
 services:
   - docker
 
-before_install:
-  - bundle install
+before_script:
   - bundle exec rake build:all
 
 script:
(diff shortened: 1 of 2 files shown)


keymon:
LGTM, I will merge as soon as travis finishes.

keymon:
Anyway, this PR did not have a associated card!!! No card (and no coffee) == no work :stuck_out_tongue_closed_eyes:

PR #37, opened 2016-03-14
[#115318605] Add Git container with SSH support

saliceti:
# What
Story: [Create tags for deploying to staging and prod](https://www.pivotaltracker.com/story/show/115318605)

We need to push to Git using SSH. We avoid the Git resource because we would have to expose the private key. This container is required to run Git within a task.

# How to review

* Build: `rake build:git-ssh`
* Test: `rake spec:git-ssh`
* Play:
```
docker run git-ssh git --version
docker run git-ssh ssh -V
```
# Who can review
Anyone but @mtekel or myself

### DIFF
--- a/git-ssh/Dockerfile
+++ b/git-ssh/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.3
+
+ENV PACKAGES "git=2.6.4-r0 openssh-client=7.1_p2-r0"
+
+RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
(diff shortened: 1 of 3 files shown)


keymon:
LGTM, I tested it with:

```
docker run -it --rm -v $HOME/.ssh/id_rsa:/root/.ssh/id_rsa git-ssh git clone $(git remote -v | sed -n 's/.*\(git@.*\) .*/\1/p;q')
```

and works fine.

The only thing: Can you rename `git-ssh/git_spec.rb` to `git-ssh/git-ssh_spec.rb`? for consistency.

You also will need somebody to create the dockerhub repo.

PR #38, opened 2016-03-18
[#115320097] Terraform 0.6.13

saliceti:
# What
This was created to investigate issue [AWS: Can't update certificate associated with ELB](https://github.com/hashicorp/terraform/issues/5713) in story [ELB for access to logs](https://www.pivotaltracker.com/story/show/115320097).

This has not fixed that particular issue, but the whole pipeline was tested on that terraform version and it worked fine.

There is a also a minor improvement as the Alpine version is now pinned to 3.3.

# How to review
Use `image: docker:///governmentpaas/terraform#115320097-terraform-0.6.13` in the pipeline and deploy.

# Who can review
Anyone but @HenryTK or myself

### DIFF
--- a/terraform/terraform_spec.rb
+++ b/terraform/terraform_spec.rb
@@ -44,7 +44,7 @@ describe "Terraform image" do
   it "has the expected Terraform version" do
     expect(
       command("terraform version").stdout
-    ).to include("Terraform v0.6.12")
+    ).to include("Terraform v0.6.13")
   end
 
   it "installs SSH" do
--- a/terraform/terraform_spec.rb
+++ b/terraform/terraform_spec.rb
@@ -19,8 +19,8 @@ describe "Terraform image" do
     set :docker_image, find_image_id('terraform:latest')
   }
 
-  it "installs the Alpine linux" do
-    expect(file("/etc/alpine-release")).to be_file
+  it "installs Alpine" do
+    expect(command("cat /etc/issue | head -1").stdout).to include("Alpine Linux")
   end
 
   it "installs Root Certificates" do
--- a/git-ssh/Dockerfile
+++ b/git-ssh/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.3
 
-ENV PACKAGES "git=2.6.4-r0 openssh-client=7.1_p2-r0"
+ENV PACKAGES "git openssh-client"
 
 RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
--- a/git-ssh/git-ssh_spec.rb
+++ b/git-ssh/git-ssh_spec.rb
@@ -4,19 +4,15 @@ require 'serverspec'
 
 PACKAGES = "git openssh-client ca-certificates"
 GIT_VERSION = "2.6.4"
-OPENSSH_VERSION = "7.1p2"
+OPENSSH_VERSION = "7.2p2"
 
 describe "image" do
   before(:all) {
     set :docker_image, find_image_id('git-ssh:latest')
   }
 
-  it "installs the right version of Alpine" do
-    expect(os_version).to include("Alpine Linux 3.3")
-  end
-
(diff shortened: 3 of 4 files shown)


terraform/terraform_spec.rb
```diff
@@ -23,6 +23,14 @@
     expect(file("/etc/alpine-release")).to be_file
   end
 
+  it "installs the right version of Alpine" do
+    expect(os_version).to include("Alpine Linux 3.3")
+  end
```
alext: I'm not sure I see the value in testing this. Do we care what version of alpine this is? All we care about is that the container has the necessary tools, and that they work.

alext:
It would be good to also update the version of terraform installed in the paas-cf Travis.yml file so that the unit tests use the same version.

mtekel:
This version has a lot of good fixes, e.g.:
* provider/aws: aws_instance now allows changes to security groups without force new resource (#5193)
* core: Includes upstream HCL fix to properly detect unbalanced braces and throw an error (#5400)

dcarley:
The build is failing for some reason that I don't understand:
```
Step 2 : RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
 ---> Running in 8550d96c7efd
fetch http://dl-cdn.alpinelinux.org/alpine/v3.3/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.3/community/x86_64/APKINDEX.tar.gz
ERROR: unsatisfiable constraints:
  openssh-client-7.2_p2-r0:
    breaks: world[openssh-client=7.1_p2-r0]
The command '/bin/sh -c apk add --update $PACKAGES && rm -rf /var/cache/apk/*' returned a non-zero code: 1
```

@saliceti could you look at this and address the comments. I'm :+1: for upgrading

mtekel:
There is some indication that 0.6.13 might fix the "flapping" NAT gateway route table updates, see latest comments in https://github.com/hashicorp/terraform/issues/4097#issuecomment-198567264

terraform/terraform_spec.rb
saliceti: I feel more comfortable pinning the versions of dependencies so upgrading terraform version is a minimal change, instead or risking unexpected behaviour due to dependencies.

saliceti:
The build issue was due to the `git-ssh` container. It depends on `openssh-client` package, but the version it relies upon was removed from the index. Bumping the version fixes the build.

It questions wether we should pin versions at all as it looks like Alpine doesn't keep previous versions.

terraform/terraform_spec.rb
dcarley: I'm :+1: for pinning the version, but :-1: for testing the specific version.

Jonty: If we stop checking the Alpine version should we also remove the Terraform version check on line 55?

combor: I'm not sure what's the value of pinning the version and running tests that do not check it?
If we want to be sure that terraform is in specific version as it contains fixes that we want then we should check if this specific version was installed. And in my opinion the same applies to any other software that we use including OS.

dcarley: > If we stop checking the Alpine version should we also remove the Terraform version check on line 55?

We shouldn't remove it, because we care about the specific version of Terraform and have written code in the `Dockerfile` to use the version in the environment variable. We'd want to know if (hypothetically) someone changed the URL to "latest" (if that exists) or the Terrraform website provided the wrong version.

Colin and I talked a bit about testing the Alpine version. I think it makes more sense to test that it's Alpine (because we want a minimal container and ability to use `apk` commands) rather than the specific version (which we want for pinning but not because of that specific version).

Though there's a separate can of worms about whether we should be pinning the versions of packages within an Alpine series.

saliceti:
Comments addressed

PR #39, opened 2016-03-22
[#115624569] Terraform container includes awscli tools

keymon:
What?
-----

In order to workaround a some missing features in terraform:

https://github.com/hashicorp/terraform/issues/5778

we want to be able to use a `local_exec` provisioner calling awscli commands.

We don't want to do it in a different task, to avoid spread the configuration
across different files and components. Using awscli from terraform
allows us keep all the related configuration is contained in the terraform
config files.

How to test this?
----------------

```
rake build:terraform spec:teraform
```

Travis should pass.

you can run `docker run -it terraform aws help`

Who?
----

Anyone but @keymon

### DIFF
--- a/terraform/Dockerfile
+++ b/terraform/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.3
+FROM governmentpaas/awscli
 
 ENV PATH $PATH:/usr/local/bin
 ENV TERRAFORM_VER 0.6.13


dcarley:
I'm uncomfortable about this. Can you link to the code (even if it's going to change or be rebased) that uses this?

keymon:
Here, in general in all that branch https://github.com/alphagov/paas-cf/commit/b2af22bc81a5be7415e5db0233d08d7c17511fa1

keymon:
After discussing this we will not include awscli in terraform. Instead we will update the container to use the master version of terraform and split the execution in different tasks as needed.

PR #40, opened 2016-03-22
[#115624569] Create the /root/.ssh directory and configure git email and user

keymon:
what?
----

To initialise the concourse pool resource, we need to do some git commits.

For that We need to update the known_hosts file and configure git user and mail.

How to test?
-----------

```
rake build:git-ssh spec:git-ssh
```

And travis should pass

who?
----

Anyone but @keymon


mtekel:
works as described

PR #14, opened 2016-01-20
[#108441228] Add container for AWS cli tool

keymon:
[#108441228 Automatically shutdown dev environments at night / weekends](https://www.pivotaltracker.com/story/show/108441228)

What
----

We want to be able to stop the vagrant AWS vm at nights. We will use AWS
cli tools to send the command to terminate it.

Scope
-----

Add a container including the AWS client tools. We create it based on
the curl container, so that we can use curl for querying the metadata.

How to test this?
-----------------

Run the tests:

```
bundle install # I recommend rvm
rake build:awscli spec:awscli
```

Run the container:

```
docker run awscli aws --version
```

Who?
----
Anyone but @keymon or @actionjack

PR #43, opened 2016-03-24
[#115624569] Instal awscli in self-update-pipelines container

keymon:
[#115624569 Implement locking for pipelines](https://www.pivotaltracker.com/story/show/115624569)

What?
---

For the setup of the pool resource for the pipeline locking, we need
to download some objects from S3. We decided use aws cli for this.

Context?
------

In this commit we install awscli from pip in ubuntu.

Test this?
-------

check #14 and #35

Who?
---

Anyone but @keymon

### DIFF
--- a/self-update-pipelines/self-update-pipelines_spec.rb
+++ b/self-update-pipelines/self-update-pipelines_spec.rb
@@ -2,6 +2,9 @@ require 'spec_helper'
 require 'docker'
 require 'serverspec'
 
+AWSCLI_BIN = "/usr/local/bin/aws"
+AWSCLI_VERSION = "1.9.21"
+
 describe "self-update-pipelines image" do
   before(:all) {
     set :docker_image, find_image_id('self-update-pipelines:latest')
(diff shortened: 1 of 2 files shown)


self-update-pipelines/self-update-pipelines_spec.rb
```diff
@@ -31,4 +34,22 @@
...
+
+  it "has the 'aws' version #{AWSCLI_VERSION}" do
+    def awscli_version
+      command("aws --version").stderr.strip
+    end
+    expect(awscli_version).to match(/aws-cli\/#{AWSCLI_VERSION} /)
+  end
+
+  it "the 'aws help' command works" do
+    expect(command("aws help").stdout).to match(/Optional  parameters/)
```
dcarley: The whitespace in the `match()` looks odd, which makes me wonder if this test could be brittle. I think it's sufficient to test that `aws --version` works.

keymon: We want to test if `aws help` actually works, because it requires having a functional groff and lesser.

I will pick a different text to check it then.

saliceti:
The commit message could mention why we install groff and less.

saliceti:
Tested OK

PR #44, opened 2016-04-08
[#117021589]  Go version bump to 1.6

combor:
# What
Current [vegeta](github.com/tsenart/vegeta) code has been moved to Go 1.6 and fails with earlier Go versions with:
```
 github.com/tsenart/vegeta/lib/attack.go:156: tr.TLSNextProto undefined (type *http.Transport has no field or method TLSNextProto)
```
We need to bump GO version to 1.6

# How to review

run:
```
docker build -t cf-acceptance-tests .
docker run cf-acceptance-tests go version
``` 
and check if go version is 1.6

# Who can review 

Not @keymon or @combor

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.5.3-wheezy
+FROM golang:1.6.0-wheezy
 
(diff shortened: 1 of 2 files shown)


alext:
Looks good to me.  Would be good to squash these 2 commits together.

combor:
Done.

mtekel:
works fine

PR #47, opened 2016-04-14
[#117522857] Install bash in the git-ssh container

keymon:
What
----

We use fair complex scripts with this containers, on which we want
to set options like pipefail [1]. Alpine `sh` does support that option
but not the GNU sh, which causes the scripts fail if they are
executed or tested outside of this container (e.g. travis).

To solve this, we install bash in the container so that we can use
it in the scripts and make them portable.

How to test
---------

just run `rake build:git-ssh spec:git-ssh`. Travis should succed

Who?
---

Anyone but @keymon or @henrytk

[1] https://bclary.com/blog/2006/07/20/pipefail-testing-pipeline-exit-codes/

### DIFF
--- a/git-ssh/Dockerfile
+++ b/git-ssh/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.3
 
-ENV PACKAGES "git openssh-client gnupg"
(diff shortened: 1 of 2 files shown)


saliceti:
Works as designed.

PR #49, opened 2016-04-29
[#116230113] Upgrade terraform to 0.6.15

dcarley:
## What

- Use binary release of Terraform instead of building from source
- Validate the checksum of the release we download
- Upgrade to Terraform 0.6.15 now that the fixes we need are released

Please see the individual commits for more detail.

## How to review

You'll want to test that the Deployer Concourse pipelines in [alphagov/paas-cf](https://github.com/alphagov/paas-cf) still work by:

- Temporarily modify your pipelines to use the branch build of the Docker container:

    ```
gsed -ri 's|(governmentpaas/terraform)|\1#upgrade_terraform_0.6.15|' concourse/pipelines/*.yml
```
- Upload the modified pipelines:

    ```
make dev pipelines
```
- Run the pipeline from end-to-end and confirm that all of the Terraform jobs are green.

## Who can review

Not @dcarley.


alext:
A nice detail with the new version is some improved output for long-running operations eg:
```
aws_db_instance.cf: Still destroying... (2m50s elapsed)
```

PR #51, opened 2016-05-05
We use Go 1.6 not 1.5

combor:
# What
The README was wrong when it comes to Go complier version.

# How

Check it the version in Dockerfile and Readme are same.

# Who

Not me

### DIFF
--- a/cf-acceptance-tests/README.md
+++ b/cf-acceptance-tests/README.md
@@ -3,7 +3,7 @@ Container for running CloudFoundry acceptance tests.
 It includes all the dependencies of the [CF test
 helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers):
 
-* Go 1.5 compiler
+* Go compiler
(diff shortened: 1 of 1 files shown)


cf-acceptance-tests/README.md
```diff
@@ -3,7 +3,7 @@ Container for running CloudFoundry acceptance tests.
 It includes all the dependencies of the [CF test
 helpers](https://github.com/cloudfoundry-incubator/cf-test-helpers):
 
-* Go 1.5 compiler
+* Go 1.6 compiler
```
dcarley: Can we omit the version altogether? That would prevent us forgetting to update it again in the future.

combor: Sure, even better.

dcarley:
Thanks for updating and fixing. I'm not going to wait for the tests to re-run, since it's only a documentation change.

PR #52, opened 2016-05-09
[#117512371] Upgrade the cached bosh-aws-cpi to version 52

alext:
# What

As part of the RDS broker story, we needed to upgrade the AWS CPI to get the change to allow security groups to be defined in a resource pool (https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/pull/28).

This PR therefore updates the cached CPI compilation to match the version used in the pipelines.

🚨  This should be merged along with https://github.com/alphagov/paas-cf/pull/260

# How to review

Code review.  Build container, and verify that tests pass.

# Who can review

Anyone but myself, @timmow or @jimconner

### DIFF
--- a/bosh-init/Dockerfile
+++ b/bosh-init/Dockerfile
@@ -13,8 +13,8 @@ RUN apt-get install -y wget
 RUN wget -q -O /usr/local/bin/bosh-init --no-check-certificate ${BOSH_INIT_URL}
 RUN chmod +x /usr/local/bin/bosh-init
 
-ENV BOSH_AWS_CPI_URL https://bosh.io/d/github.com/cloudfoundry-incubator/bosh-aws-cpi-release?v=44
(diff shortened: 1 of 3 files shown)


mtekel:
Should we also update BOSH CLI version? There have been some updates to the client, the latest version is 1.3232.

mtekel:
Also, we should upgrade deployer concourse and BOSH CPIs to v52 as well. We should create a story for that.

alext:
I'm not sure what you mean about the BOSH CPI.  It's upgraded in the linked PR above.

alext:
> Should we also update BOSH CLI version?

Perhaps, but not as part of this story. We're upgrading the CPI for a specific reason in alphagov/paas-cf#260, and this is just to keep things in line.

PR #53, opened 2016-05-16
[#117512371] Add ruby to cf-cli container.

alext:
# What

This adds ruby to the cf-cli container so that we can script operations against the CF API more easily (eg
adding the security groups defined in the manifest).

# How to review

Verify that the tests pass. Verify that the cf-cli container does indeed have ruby available.

# Who can review

Anyone but myself.

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:wheezy
+FROM ruby:2.2-slim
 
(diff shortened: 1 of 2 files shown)


dcarley:
Alex and I talked about this and agreed that it's the pragmatic thing to do, since both quotas and security groups follow this pattern and it's harder to read in Concourse.

PR #54, opened 2016-05-26
[#118838943] Update bosh_cli gem to 1.3232.0

saliceti:
# What
bosh_cli gem version is now pinned in paas-cf. We want to use the same
version in the container as well.

# How to review
* Build and run local tests
* Run the bosh-cf pipeline with the new container by replacing `docker:///governmentpaas/bosh-cli` with `docker:///governmentpaas/bosh-cli#update_bosh_cli_gem`

# Who can review
Anyone but me


jimconner:
Looks good and works well for me. Hitting the big green merge button.

PR #55, opened 2016-06-15
Bump AWS CLI version to 1.10.38

mtekel:
# What
In order to add aws sts `get-caller-identity` command which we need for
tagging RDS instances in upgrade master password story (#118841005).

# Testing
Run `rake build:awscli`. Then `rspec awscli`. You can check that the version inside container matches by hand and you can also check that it has `aws sts get-caller-identity` command.

# Who
not me


combor:
I was able to build container, rspec finished without any failures and `aws sts get-caller-identity` works

PR #56, opened 2016-06-16
[#111700216] Upgrade spruce to 1.4.5

mtekel:
### What
Spruce now provides static builds, which enables us to run it from
Alpine (see https://github.com/geofffranks/spruce/issues/128).

Use new spruce binary to reduce container size by 80%:
 * old 73MB
 * new 14MB

### Testing
Render manifests using old spruce. Render manifest using new spruce. Compare - there should be no differences. You could also upgrade your local spruce to 1.4.5 & run manifest rendering tests (or all tests - `make test`)

### Who
not @mtekel

PR #57, opened 2016-06-22
[#111700216] Upgrade spruce to 1.5

mtekel:
### What
Spruce now provides static builds, which enables us to run it from
Alpine (see https://github.com/geofffranks/spruce/issues/128).

Use new spruce binary to reduce container size by 80%:
 * old 73MB
 * new 14MB

### Testing
Render manifests using old spruce. Render manifest using new spruce. Compare - there should be no differences. You could also upgrade your local spruce to 1.4.5 & run manifest rendering tests (or all tests - `make test`)

You can also render locally and compare outputs with previous spruce version:
```
cd manifests
shared/build_manifest.sh cf-manifest/manifest/*.yml cf-manifest/manifest/data/*.yml cf-manifest/manifest/env-specific/cf-default.yml cf-manifest/manifest/runtime/runtime.yml shared/spec/fixtures/*.yml shared/spec/fixtures/terraform/terraform-outputs.yml >manifest.yml
```

### Who
not @mtekel

### DIFF
--- a/spruce/Dockerfile
+++ b/spruce/Dockerfile
@@ -1,12 +1,10 @@
-FROM golang:1.5.1-alpine
+FROM alpine:3.4
 
-ENV SPRUCE_VERSION 0.13.0
+ENV SPRUCE_VERSION 1.5.0
 
-RUN apk add --update git \
-  && go get -d github.com/geofffranks/spruce \
-  && cd ${GOPATH}/src/github.com/geofffranks/spruce \
-  && git checkout v${SPRUCE_VERSION} \
-  && export GOPATH=$(pwd)/Godeps/_workspace:$GOPATH \
-  && go install \
-  && apk del git \
+RUN apk add --update wget ca-certificates \
+  && wget https://github.com/geofffranks/spruce/releases/download/v${SPRUCE_VERSION}/spruce-linux-amd64 \
+  && chmod +x spruce-linux-amd64 \
+  && mv spruce-linux-amd64 /usr/local/bin/spruce \
+  && apk del wget ca-certificates \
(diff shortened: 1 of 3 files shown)


combor:
All local tests were done here: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/56

I'm going to deploy from this branch now.

combor:
We need to change README as we mention that this container is using go alipne which is not anymore.

mtekel:
updated

combor:
Looks good. The only changes discovered by bosh were:
```
Detecting deployment changes
----------------------------
jobs:
- name: rds_broker
  properties:
    rds-broker:
      master_password_seed: "<redacted>"
      master_password_seed: "<redacted>"
      state_encryption_key: "<redacted>"
      state_encryption_key: "<redacted>"
```

Merging...

dcarley:
> Looks good. The only changes discovered by bosh were:
> 
> ```
> Detecting deployment changes
> ----------------------------
> jobs:
> - name: rds_broker
>   properties:
>     rds-broker:
>       master_password_seed: "<redacted>"
>       master_password_seed: "<redacted>"
>       state_encryption_key: "<redacted>"
>       state_encryption_key: "<redacted>"
> ```

@combor can you explain why that was expected?

combor:
New deployment of rds-broker so cf-secrets have two new lines:
```
diff ~/Downloads/cf-secrets-before.yml ~/Downloads/cf-secrets-after.yml
30a31,32
>   rds_broker_master_password_seed: <redacted>
>   rds_broker_state_encryption_key: <redacted>
```

PR #59, opened 2016-07-08
[#122183539] Fix wget in bosh-cli container for https sites.

alext:
# What

The current version of the ruby:2.2-alpine no longer includes the
openssl CLI. That means that the busybox provided wget command can no
longer connect to https sites, and instead gives the following error:

```
wget: can't execute 'ssl_helper': No such file or directory
wget: error getting response: Connection reset by peer
```

This adds openssl to this container to fix this issue.

This is a partial revert of cfc7785e, which updated the tests to check for
the libssl version instead because we didn't think that the openssl CLI
was needed.

## How to review

Verify that the tests pass in Travis.

Verify that wget works with https sites:

```sh
cd bosh-cli
docker rmi ruby:2.2-alpine # To ensure we get the latest.
docker build -t bosh-cli .
docker run bosh-cli wget https://www.google.com
```

## Who can review

Anyone but myself.

### DIFF
--- a/bosh-cli/Dockerfile
+++ b/bosh-cli/Dockerfile
@@ -2,4 +2,4 @@ FROM ruby:2.2-alpine
 
 RUN gem install bosh_cli -v 1.3232.0 --no-rdoc --no-ri
 
-RUN apk add --update openssh-client file && rm -rf /var/cache/apk/*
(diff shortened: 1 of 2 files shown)


alext:
I considered adding a test that did a wget to something like https://www.google.com, but I wasn't sure it was a good idea to rely on an external site in the tests.

keymon:
:+1:

PR #60, opened 2016-07-25
[#117991263] Upgrade bosh-init to 0.0.95

mtekel:
### What

Upgrade bosh-init version. This is required to deploy new concourse version, which uses various
"BOSH 2.0" features that are not supported in our previous bosh-init version.

### Testing
As a part of https://github.com/alphagov/paas-cf/pull/381

### Who
not @mtekel or @combor

### DIFF
--- a/bosh-init/Dockerfile
+++ b/bosh-init/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:14.04
 
-ENV BOSH_INIT_VERSION 0.0.80
+ENV BOSH_INIT_VERSION 0.0.95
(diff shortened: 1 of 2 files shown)


alext:
Is the new version backwards-compatibile? Will it work with our existing pipelines/manifests?

mtekel:
We only use bosh-init to deploy concourse and bosh micro. This works with concourse, we will test bosh-init as a part of alphagov/paas-cf#381

mtekel:
As a side note, this container is quite beefy (517MB!), maybe we could spend some time reducing the size...

PR #61, opened 2016-08-17
[#126965751] Add jq 1.5 to the cf-cli container

keymon:
[#126965751 Test users not removed from CC DB](https://www.pivotaltracker.com/story/show/126965751)

What
====

We need to parse the output of `cf curl /v2/users` to list all the existing users created by the acceptance tests. For this jq is perfect.

We consider that this really useful and it will be used in other stories in the future, so we add it as a basic tool in the cf-cli container.

We want to use the regex based functions, so jq must be >= 1.5 and must be compiled with ONIGURUMA regex library. The Ubuntu package is 1.4, so we will pull the static binary released upstream.

Dependencies
-----------------

Please merge #65 first and rebase.

This could be reviewed together with https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/62

How to test?
-----------

Travis shall pass.

Then you can test: `docker run -ti cf-cli jq --version`

Who?
----

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -6,4 +6,4 @@ RUN apt-get update \
       && apt-get install -y --no-install-recommends $PACKAGES \
       && rm -rf /var/lib/apt/lists/*
 RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.16.1' | tar -zx -C /usr/local/bin
-
+RUN curl -L 'https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64' -o /usr/local/bin/jq && chmod +x /usr/local/bin/jq
--- a/cf-cli/cf-acceptance-tests_spec.rb
+++ b/cf-cli/cf-acceptance-tests_spec.rb
@@ -33,6 +33,12 @@ describe "cf-cli image" do
     ).to eq(0)
   end
 
+  it "has jq available" do
+    cmd = command("jq --version")
+    expect(cmd.exit_status).to eq(0)
+    expect(cmd.stdout).to match(/^jq-1.5/)
+  end
+
   it "has ruby 2.2 available" do
     cmd = command("ruby -v")
     expect(cmd.exit_status).to eq(0)


dcarley:
Build failed to install `json` gem. I haven't investigated why.

I'm not totally against this change, but I'm unsure about the precedent it might set for shell scripts manipulating data that could benefit from being written in another language that supports data structures. Could you use Ruby (which is in the `cf-cli` container) to achieve what you're doing at the moment? Would it be much harder?

richardc:
An equivalent ruby script is possible, but would be another small piece of software we need to write and maintain, something like:

```
document = JSON.load(STDIN.read)
document['resources'].each do |elem|
   username = elem["entity"]["username"]
   next unless username =~ /"^custom-acceptance-test-user/
   puts username
end
```

vs.

```
jq -r '.resources[] | .entity.username | select(. != null) | select(test("^custom-acceptance-test-user"))'
```

`jq` is approaching a the place of a standard 'awk for json' tool so I think it's safe to adopt.

richardc:
Travis build is failing as the ruby in use has switched from [2.2.3p173](https://travis-ci.org/alphagov/paas-docker-cloudfoundry-tools/jobs/147159510#L173) to [2.3.1p112](https://travis-ci.org/alphagov/paas-docker-cloudfoundry-tools/builds/152936129#L242).

2.3.1 is api incompatoible with the pinned version of the json gem.  Recommend pinning ruby version in the .travis.yml

keymon:
Rebased after merging #65

benhyland:
@keymon I need to run bundle install to follow the manual testing instructions, and I can't use rbenv without a .ruby-version. I'm happy to add that as part of this PR, but what version should I use?

keymon:
I guess your comment refers to https://github.com/alphagov/paas-cf/pull/430

I think we should discuss within the team if we should add `.ruby-version` in the repos. About version, I think we agreed on 2.2

benhyland:
@keymon my comment refers to the instruction `You can test locally with rake build:cf-cli spec:cf-cli.` in this PR.
Pinning the Ruby version is not really optional, but very happy to discuss tooling consistency within the team. I'll add a `.ruby-version` at 2.2 in the meantime. Thanks.

keymon:
Alpine 3.3 uses ruby 2.3, instead of 2.2

I did not want to change such stuff.

mtekel:
You mean 3.4 uses 2.3? and 3.3 2.2?

mtekel:
builds and tests correctly with ruby 2.2.2. I have also pulled image built by docker hub and that one also looks OK, merging

mtekel:
the comments about versions belong to https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/62

PR #62, opened 2016-08-17
[#126965751] Use Alpine for cf-cli container base image

keymon:

What
====

We use alpine base image whenever is possible to speed up the pipelines.   But in the case of cf-cli we were using Ubuntu because [the cf-cli was not compiled statically](https://github.com/cloudfoundry/cli/issues/807). Since 6.17.1. they build it statically, so we can move the container to alpine.

Dependencies
------------

Please merge #65, and #61 first and rebase

How to test
-----------

Travis shall pass.

Then you can test: `docker run -ti cf-cli jq --version`

Who?
----

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -1,9 +1,8 @@
-FROM ruby:2.2-slim
+FROM alpine:3.3
 
-ENV PACKAGES unzip curl ca-certificates git
+ENV PACKAGES "unzip curl openssl ca-certificates git ruby"
 
-RUN apt-get update \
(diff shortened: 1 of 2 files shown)


cf-cli/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM ruby:2.2-slim
+FROM alpine:3.3
```
mtekel: why not 3.4?

keymon: Alpine 3.4 uses ruby 2.3, instead of 2.2

I did not want to change such stuff.

mtekel:
works fine (in the pipeline)

PR #65, opened 2016-08-17
[#126965751]  Upgrade to docker-api 1.31 gem and pin ruby to 2.2 to support travis

keymon:

What
====

Pin ruby version in travis to ruby 2.2, as  we are having issues building json gem in travis ([here](https://travis-ci.org/alphagov/paas-docker-cloudfoundry-tools/builds/147760906#L165) vs [here](https://travis-ci.org/alphagov/paas-docker-cloudfoundry-tools/builds/152935537#L301)) and we decided to use ruby 2.2 as the standard in the team.

Additionally Travis docker service has been upgraded and we need to upgrade the clients. Otherwise, the builds fail [with this error](https://travis-ci.org/alphagov/paas-docker-cloudfoundry-tools/builds/153019275#L1860):
``` 
Failure/Error: image = Docker::Image.all().detect{ |i| i.info['RepoTags'].include? name }
Docker::Error::ClientError:
400 Bad Request: malformed Host header
```

How to test
-----------

Travis shall pass.

Who?
----

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+language: ruby
+rvm:
+  - 2.2
+
 sudo: required
 
 services:
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 
 source 'https://rubygems.org'
 
-gem 'docker-api', '~> 1.21'
+gem 'docker-api', '~> 1.31'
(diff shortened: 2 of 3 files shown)


keymon:
Thx for merging, I will update the other branches.

PR #67, opened 2016-08-18
[#126965751] Install ruby gem json in cf-cli container

keymon:

What
====

After merging #62, our pipeline fail when executing a ruby script which required json.                                                            
                                                                                
Although json is one of the basic packages shipped with ruby[1], Alpine extracts it in a different package ruby-json. We need to explicitely add that dependency.                                                
                                                                                
We avoid running `bundle install` in the pipeline to avoid increase the time to run. We normally try to keep the scripts simple and without dependencies, but `json` is a really basic library.                 
                                                                                
[1] https://zzz.buzz/2016/02/29/default-gems-bundled-with-ruby/                 
[2] http://git.alpinelinux.org/cgit/aports/tree/main/ruby/APKBUILD?h=3.4-stable&id=cbdb9b2e42bbd7a67f0945bf7bcb8e79ddcae742#n190
                                                                                

How to test
-----------

Travis shall pass.

Who?
---

Anyone but @keymon

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.3
 
-ENV PACKAGES "unzip curl openssl ca-certificates git ruby"
+ENV PACKAGES "unzip curl openssl ca-certificates git ruby ruby-json"
 
 RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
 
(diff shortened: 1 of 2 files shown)


bleach:
`json` is a [default gem that ships with ruby](https://zzz.buzz/2016/02/29/default-gems-bundled-with-ruby/). Why do we even need to install it? Does alpine remove the default gems?

bleach:
Answering my own question:

Assuming [this](http://git.alpinelinux.org/cgit/aports/tree/main/ruby/APKBUILD?h=3.4-stable&id=cbdb9b2e42bbd7a67f0945bf7bcb8e79ddcae742) is the correct APKBUILD it takes `json`, `irb`, `rake` etc out of the main package and puts them into subpackages.

keymon:
image

HenryTK:
Travis looks good, running locally it builds and tests pass, and tasks in the pipeline which use it are running okay. Merging.

PR #68, opened 2016-08-30
[#123003931] Add concourse `enum` resource type

richardc:
## What

[#123003931](https://www.pivotaltracker.com/n/projects/1275640/stories/123003931) Improve "User" (PaaS Operator) Experience for creating  new Environment

Introduces a new concourse enum resource type.

This type enables us to model the state of an enum backed in a s3 bucket.  Its key improvements are to encapsulate a pattern we found useful while implementing [#123003931](https://www.pivotaltracker.com/n/projects/1275640/stories/123003931)

## How to review

* Checkout this branch and modify the example pipeline to point at the docker tag for this branch:

~~~
diff --git a/enum-resource/example_pipeline.yml b/enum-resource/example_pipeline.yml
index 5b0b624..ad496f5 100644
--- a/enum-resource/example_pipeline.yml
+++ b/enum-resource/example_pipeline.yml
@@ -4,6 +4,7 @@ resource_types:
   type: docker-image
   source:
     repository: governmentpaas/enum-resource
+    tag: feature_123003931_concourse_enum_type

 resources:
 - name: sample-enum
~~~

* Upload the pipeline to a development concourse

~~~
fly set-pipeline -p example-enum -c example_pipeline.yml -v bucket-name=${DEPLOY_ENV}-state -v aws_region=eu-west-1
~~~

* Run the jobs `sample_enum_get_job` and `sample_enum_set_job` and see the enum flips from a default value (`no`) to a set value (`yes`)
* You can also check against the state in the s3 bucket

~~~
aws s3 cp s3://${DEPLOY_ENV}-state/sample-enum -
~~~

## Who can review

Not @richardc @benhyland


alext:
I'm concerned that this doesn't include any tests for the functionality of this resource.

enum-resource/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+    ln /opt/resource/smuggler /opt/resource/out
+
+# Add the config file for this resource
+ADD ./smuggler.yml /opt/resource/
+
+# Add shared code used in smuggler.yml
+ADD ./*.sh /opt/resource/
+
+# Remove build dependencies
+RUN apk del ${BUILD_PACKAGES} && rm -rf /var/cache/apk/*
```
combor: AFAIR this will just create another layer with removed cache and in the end will not reduce the size of the container. Same for line below.

enum-resource/smuggler.yml
```diff
@@ -0,0 +1,36 @@
+---
+commands:
+  check: |
+    /opt/resource/wrapped/s3/check > response.json
+    # If it is the first run, just dispatch a - string to for 'in' to be triggered
+    jq 'if . == [] then [{ "version_id":"-"}] else . end' < response.json
```
keymon: In CI it seems that we got the `-` after getting a normal version https://deployer.master.ci.cloudpipeline.digital/pipelines/create-bosh-cloudfoundry/resources/healthcheck-deployed

commenting this with @alext, maybe the issue is that the S3 resource returns a empty list if the current version is the same that is passed as a argument?

We should test that and maybe add a different check. Maybe based on `$SMUGGLER_VERSION_ID`? 

Actually, IIRC S3 check fails if the object is not present.

keymon: Yes,  I have checked that is the case. I hijacked the container in CI (`./bin/fly -t master hijack -c create-bosh-cloudfoundry/healthcheck-deployed sh`) and run

``` 
# /opt/resource/wrapped/s3/check <<EOF
{
  "source": {
    "bucket": "master-state",
    "region_name": "eu-west-1",
    "versioned_file": "healthcheck-deployed",
    "valid_values": ["yes", "no"],
    "default_values": ["no"]
  },
  "version": { "version_id":"4lO7GqPDXFewTWypE8cOkeZdcSxWYiW3" }
}
EOF 
[]
```

It does return empty list `[]`.

The same if you specify a missing file:

```

``` 
# /opt/resource/wrapped/s3/check <<EOF
{
  "source": {
    "bucket": "master-state",
    "region_name": "eu-west-1",
    "versioned_file": "something_random",
    "valid_values": ["yes", "no"],
    "default_values": ["no"]
  },
  "version": { "version_id":"4lO7GqPDXFewTWypE8cOkeZdcSxWYiW3" }
}
EOF 
[]
```

Comment (keymon) [inline enum-resource/smuggler.yml]: Maybe the right implementation can be (I did not test it)

```
if [ "$SMUGGLER_VERSION_ID" == "" ]; then
   echo '[{ "version_id":"-"}]'
else 
   /opt/resource/wrapped/s3/check
fi
```

PR #70, opened 2016-09-09
[#130003999] Remove container size info from terraform README.

alext:
## What

This no longer applies since we've upgraded to 0.7 (in fa630ad) as
terraform now has a single binary that includes all the built-in
providers.

## How to review

Review update README and confirm it makes sense to remove these details.

## Who can review

Anyone but myself.


mtekel:
Indeed.

PR #74, opened 2016-10-11
[#132060265] Add git to bosh-cli image

keymon:
[#132060265 Support: Wrong version of aws-broker deployed in CI](https://www.pivotaltracker.com/story/show/132060265)

What?
-----

We want to add a check to verify that the tag of the cloned version of  a bosh release matches with the intended version to upload before uploading it.

For that, we need git in this image to check tags or gather any git information from the git repositories of the bosh releases that we build.

How to review?
--------------

Travis should pass. You can run `rake build:bosh-cli spec:bosh-cli`

You can review it with the corresponding PR in paas-cf.

Who?
----

Anyone but @keymon

### DIFF
--- a/bosh-cli/Dockerfile
+++ b/bosh-cli/Dockerfile
@@ -2,4 +2,4 @@ FROM ruby:2.2-alpine
 
 RUN gem install bosh_cli -v 1.3232.0 --no-rdoc --no-ri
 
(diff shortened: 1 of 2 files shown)


richardc:
Looks good and testing the feature in the linked PR behaves as expected.  Merging.

PR #75, opened 2016-10-20
[#132651425] Upgrade cf-cli in cf-acceptance-tests to 6.21.1

alext:
## What

CF 245 requires a newer CLI version than the one in this container (6.16.1). This PR therefore upgrades this to 6.21.1, which is also the version that the `cf-cli` container is already using (it was upgraded in a2d25c4).

## How to review

To test locally:

`bundle exec rake build:cf-acceptance-tests spec:cf-acceptance-tests`

`docker run --rm cf-acceptance-tests cf -v` should return version 6.21.1

## Who can review

Anyone but @keymon or myself.

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,6 +1,6 @@
 FROM golang:1.6.0-wheezy
 
-RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.16.1' | tar -zx -C /usr/local/bin
+RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.21.1' | tar -zx -C /usr/local/bin
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


mtekel:
Works. Also, it looks like this could be built from alpine, golang has official alpine base image: https://hub.docker.com/_/golang/ . Only additional package is unzip, which I believe would be available for alpine...

btw. for cf -v, I get `cf version 6.21.1+6fd3c9f-2016-08-10` - is that what you expect?

alext:
> Also, it looks like this could be built from alpine

It looks like it could. I think that's something for a separate PR though, as that may require more extensive testing with the acceptance tests etc..

> btw. for cf -v, I get cf version 6.21.1+6fd3c9f-2016-08-10 - is that what you expect?

Yes, that matches what I get.

mtekel:
Story: https://www.pivotaltracker.com/n/projects/1275640/stories/132820493

PR #76, opened 2016-10-27
[#132681767] Upgrade the cached bosh-aws-cpi to version 60

keymon:
[#132681767 Upgrade BOSH #2: stemcell, CPI and new features](https://www.pivotaltracker.com/story/show/132681767)

What
----

We upgraded the CPI in bosh in PR https://github.com/alphagov/paas-cf/pull/570

We upgrade the cached CPI in the bosh-init to bring it inline with the version used in our pipelines.

Dependencies
------------

This should be merged together with PR https://github.com/alphagov/paas-cf/pull/570

How to test?
-----------

Travis tests this, should pass.

Otherwise: `rake build:bosh-cli spec:bosh-cli`

Who?
----

Anyone but @keymon


mtekel:
This works, tested with https://github.com/alphagov/paas-cf/pull/570

PR #77, opened 2016-11-15
[#133751193] Upgrade terraform container to 0.7.10

keymon:
[#133751193 Enable datadog BOSH healthmonitor integration](https://www.pivotaltracker.com/story/show/133751193)

What?
-----

We want to upgrade terraform to use a patched version of the datadog provider.

Latest version is 0.7.10. There is no need to use our forked version of terraform anymore, so we go back to the code that downloads the official binary.

We were forced to upgrade the gem for docker-api for our tests, as the docker API has changed and the tests fail in the latest docker versions (~>1.12)

How to test?
-----------

Check Travis should pass.

You can optionally run `rake build:terraform spec:terraform`.

Who?
----

Anyone but @keymon or @saliceti

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 
 source 'https://rubygems.org'
 
-gem 'docker-api', '~> 1.31'
+gem 'docker-api', '~> 1.32'
 gem 'serverspec', '~> 2.19'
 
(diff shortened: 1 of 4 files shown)


mtekel:
works fine...

PR #78, opened 2016-11-16
[#132820493] cf-acceptance-tests built from alpine

mtekel:
### What

Use alpine for cf-acceptance-tests container instead of debian wheezy. This reduces size from 670MB to 360.

### Testing
To build container and run tests: `bundle exec rake build:cf-acceptance-tests` and `bundle exec rake spec:cf-acceptance-tests`. All should pass. To test new container in tasks that use it - use [test-slim-acceptance-container](https://github.com/alphagov/paas-cf/tree/test-slim-acceptance-container) branch, or cherry pick this commit: https://github.com/alphagov/paas-cf/commit/fcbafea678a645029636b38d35b4465215beff30 . Apply to your pipeline and run. Make sure to run all acceptance tests.

### Who
not @mtekel

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,9 +1,7 @@
-FROM golang:1.6.0-wheezy
+FROM golang:1.6-alpine
 
+RUN apk add --update curl unzip git bash && apk update && apk upgrade && rm -rf /var/cache/apk/*
 RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.21.1' | tar -zx -C /usr/local/bin
-RUN apt-get update \
-      && apt-get install -y --no-install-recommends unzip \
-      && rm -rf /var/lib/apt/lists/*
 
 RUN go get github.com/tools/godep
 RUN go get github.com/onsi/ginkgo/ginkgo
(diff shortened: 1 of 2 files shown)


alext:
Will this not require changes in the pipeline. At the moment we have some things that assume a Debian system (eg adding the bosh CA cert to the OS cert store).

mtekel:
We already do use alpine in many places where we do CA cert stuff (e.g. https://github.com/alphagov/paas-cf/blob/master/concourse/pipelines/create-bosh-cloudfoundry.yml#L737). This all works correctly and it's not a new thing to do it in alpine...

keymon:
The reason why we chose ubuntu wheezy  instead of alpine was that the cf-acceptance-tests are developt and tested upstream using ubuntu in the errand.

If we move to alpine, even if it works now, we are in risk that it stops working in later versions of cf-acceptance-tests because the test relies on some dependency of the base ubuntu.

Of course, you can argue that in that case we can troubleshoot it, and fix it or workaround it or revert this change. But then it is arguable the win of reducing the size of this container from 670MB to 360. Also the advantage of reducing the size in this container is low, because the tests take several minutes or hours to run, there is very little improvement here.

Because that, I suggest that we don't do this change.

Jonty:
Closing this as @bleach and @timmow have decided we shouldn't do it.

PR #79, opened 2016-11-17
[#130188061] Add bash to bosh-cli

saliceti:
# What

In order to build boshreleases that contain packages with pre_packagging steps, we need bash, as that's what bosh invokes for these scripts.[1]

[1] https://github.com/cloudfoundry/bosh/blob/stable-3262.9/bosh_cli/lib/cli/resources/package.rb#L193

# How to review
`rake build:bosh-cli spec:bosh-cli`

# Who can review
Anyone but @combor or me


mtekel:
works fine...

PR #83, opened 2017-01-23
[#135948443] Use upstream's image for CATS

henrytk:
## What

The latest Cloud Foundry upgrade (cf-release v245 to v250) means the
smoke and acceptance tests require golang 1.7+. However, the upstream golang image did not support gzip 1.6, which is used in the new acceptance test code, so we have resorted to using a Dockerfile maintained by CloudFoundry.

## Dependencies

**Warning:** merging this will break the tests until the PR: https://github.com/alphagov/paas-cf/pull/735 is not deployed in prod, because we do not pin Docker image versions. 

See notes in https://github.com/alphagov/paas-cf/pull/735 for more details.

## How to review

You can run the usual local `rake build:cf-acceptance-tests` and `rake spec:cf-acceptance-tests`.

## Who

Anyone but me or @keymon

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -8,7 +8,8 @@ RUN \
     curl \
     openssh-client \
     unzip \
-    git
+    git \
+  && rm -rf /var/lib/apt/lists/*
 
  ENV GOPATH /go
  ENV PATH /go/bin:/usr/local/go/bin:$PATH
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,9 +1,30 @@
-FROM golang:1.6.0-wheezy
+FROM ubuntu:trusty
 
-RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.21.1' | tar -zx -C /usr/local/bin
-RUN apt-get update \
-      && apt-get install -y --no-install-recommends unzip \
-      && rm -rf /var/lib/apt/lists/*
+RUN \
+  apt-get update && \
+  apt-get -y install \
+    build-essential \
+    wget \
+    curl \
+    openssh-client \
+    unzip \
+    git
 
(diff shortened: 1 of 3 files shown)


mtekel:
Out of curiosity, I have also bumped my alpine container. It was quite simple: https://github.com/alphagov/paas-docker-cloudfoundry-tools/tree/slim-acceptance-132820493

The size difference is 312MB vs 648, or reduction by 336MB/52%....

keymon:
I keep my opinion that we should use the most similar image possible. This time use the very same Dockerfile from https://github.com/cloudfoundry/cats-concourse-task/blob/master/Dockerfile makes sense.

But @dcarley was keen on using alpine, so it would be other vote towards alpine.

dcarley:
I'm not that fussed about the download size because bandwidth is relatively cheap. I have a preference for reducing the number of dependencies bundled into the base image, because it's less things to maintain, but the tradeoff here is that we'd be deliberately going against the grain of what upstream are doing.

mtekel:
My main motivation in reducing size was in relieving load on concourse, where gardnen copies full image on every run and baggageclaim has to handle much larger images. And because of continuous smoke tests, this is our most used image...

Regarding the argument about compatibility/ease of support; as it turns out, alpine was quite simple to bump (not needing any gzip trickery). When using upstream image - why don't we use that image directly but need to copy (and thus maintain) dockerfile?

keymon:
> reducing size was in relieving load on concourse

I am not sure about this. 

First lately we hadn't as many issues as before with concourse, second the issue was more about the number of containers being pulled in parallel rather than the size, and finally I think there are plenty of things we can/should do to reduce the load in concourse and will really make a difference. 

For instance we can group image containers instead of having so many different specific containers (e.g. only one container with AWS cli, terraform, bosh, ruby, python, etc.), and create custom resources rather than doing multiple data transformation using tasks and scripts.

> alpine was quite simple to bump (not needing any gzip trickery).

What about the other way round, when if we need an older version of a component?

PR #86, opened 2017-01-31
[#138486833] Terraform update to v 0.8.5

combor:
# What
As we refactored terraform config in https://github.com/alphagov/paas-cf/commit/965ae1e977c8298027399e4d09277aa531e7a05a
we started to be hitting https://github.com/hashicorp/terraform/issues/5301 again.

We need to update to a newer terraform version to successfully run terraform in our pipeline.

# How to test 

This needs to be reviewed and merged in conjunction with alphagov/paas-cf#744

# Who
Not @alext or @combor

### DIFF
--- a/terraform/Dockerfile
+++ b/terraform/Dockerfile
@@ -1,8 +1,8 @@
 FROM alpine:3.4
 
 ENV PATH $PATH:/usr/local/bin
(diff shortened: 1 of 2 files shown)


keymon:
LGTM

Issue #81, opened 2017-01-18
CF CFLI outdated


mastertinner:
Can we please rebuild the `governmentpaas/cf-cli` image? The CLI is outdated (6.21.1 whereas the latest one is 6.23.1).

combor:
Hi, thanks for reporting this. I've created PR https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/87 to update `cf cli` binary.

dcarley:
@mastertinner It's been published now: https://hub.docker.com/r/governmentpaas/cf-cli/builds/bhj5lzmkqqtkxxfp2rxm5tc/

I'm glad the images are of use to you! I'd like to note though that we consider the project as ["coding in the open" rather than "open source"](https://gds.blog.gov.uk/2012/10/12/coding-in-the-open/), so we can't guarantee that we won't make a backwards incompatible change in the future, especially as we don't currently use semantic versioning for containers.

PR #87, opened 2017-02-03
[#137881471]Update cf cli to v 6.23.1

combor:
## What
Bump version of cf-cli used in our governmentpaas/cf-cli container to
latest (currently v6.23.1).

Update brings useful features:
- Where available, error messages from 5xx responses from API endpoints now include the request id, which can be helpful for platform operators to look up details on the interaction in CF's internal logs.
- We are in the process of creating a more consistent user experience; our goal is to standardize UI output. For example, warnings and errors will consistently be outputted to stderr instead of stdout. As we iterate through the list of commands, we are also focusing on improving performance and stability.

- This release introduces commands to run, terminate and list tasks, available from CF release v247 
- `delete-space` now takes an optional org parameter to allow deletion of a space without targeting it.
- `logs` and `push` now use the doppler logging endpoint published by the CC API (/v2/info), as opposed to extracting it from the recently deprecated logging endpoint entry.
- a lot of improvements to help messages for various commands

Also we've got external request to bump the version: https://github.com/alphagov/paas-docker-cloudfoundry-tools/issues/81

## How to test

- run the create pipeline from `137881471_update_cf_cli_6_23_1` branch on `paas-cf`
- **Warning:** Important info from release notes - `For example, warnings and errors will consistently be outputted to stderr instead of stdout.` This may affect all code that is parsing `cf cli` output. We need to test it carefully. 

## Who

Not @combor

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -4,5 +4,5 @@ ENV PACKAGES "unzip curl openssl ca-certificates git ruby ruby-json"
 
 RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
 
-RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.21.1' | tar -zx -C /usr/local/bin
+RUN curl -L 'https://cli.run.pivotal.io/stable?release=linux64-binary&version=6.23.1' | tar -zx -C /usr/local/bin
 RUN curl -L 'https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64' -o /usr/local/bin/jq && chmod +x /usr/local/bin/jq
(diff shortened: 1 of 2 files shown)


dcarley:
> run the create pipeline from testing PR: -link here

🚧  pending 🚧

combor:
Fixed :)

dcarley:
@combor already checked that alphagov/paas-bootstrap doesn't use `governmentpaas/cf-cli`

I ran the `create-cloudfoundry` pipeline from alphagov/paas-cf on an environment that had been auto-deleted overnight. The whole pipeline passed, although it may not be strictly representative because there are a handful of features disabled in dev (e.g. application deployments)

So I also had a look through alphagov/paas-cf with `ack 'cf '` for places where we parse the CLI output. I couldn't find any cases where we're expecting an error or warning, except:

- https://github.com/alphagov/paas-cf/blob/e63f3f0ddceea9783afaa87384aebea47acb00c4/scripts/common.sh#L58 which still uses STDOUT
- https://github.com/alphagov/paas-cf/blob/e63f3f0ddceea9783afaa87384aebea47acb00c4/scripts/create-user.sh#L140 which redirects to STDOUT and is outside the pipeline anyway

So good to go 👍

PR #88, opened 2017-02-06
[bugfix] Remove "based on" from READMEs

dcarley:
## What

I noticed this while reviewing #87

We're not very good at maintaining the "Based on XXX image" messages in the
READMEs of each of our containers. I don't think there's much value in
providing this, since the correct information is the in the `Dockerfile` and
people should look there before deciding whether a container is appropriate
for their use.

```
➜  paas-docker-cloudfoundry-tools git:(master) for i in */README.md; do ack Based ${i} && ack FROM $(dirname $i)/Dockerfile; echo; done
Based on [alpine](https://hub.docker.com/_/alpine/) image.
FROM governmentpaas/curl-ssl

Based on the [ruby:slim](https://hub.docker.com/_/ruby/) image.
FROM ruby:2.2-alpine

Based on the [upstream image](https://hub.docker.com/r/relintdockerhubpushbot/cats-concourse-task/), which is
FROM ubuntu:trusty

Based on the [wheezy](https://hub.docker.com/_/debian/) image.
FROM alpine:3.3

Based on the [ruby:slim](https://hub.docker.com/_/ruby/) image.
FROM ruby:2.2-alpine

Based on [alpine](https://hub.docker.com/_/alpine/) image.
FROM alpine:3.3

Based on [alpine](https://hub.docker.com/_/alpine/) image.
FROM alpine:3.3

Based on the [ruby:alpine](https://hub.docker.com/r/library/ruby/tags/) image.
FROM ruby:2.3-alpine

Based on [alpine](https://hub.docker.com/_/alpine/) image.
FROM alpine:3.3

Based on the [wheezy](https://hub.docker.com/_/ruby/) image.
FROM ruby:2.2-slim
```

## How to review

Confirm that the documentation changes make sense.

## Who can review

Not @dcarley

### DIFF
--- a/awscli/README.md
+++ b/awscli/README.md
@@ -4,8 +4,6 @@ command line.
 Also curl with SSL and openssl support is included, to allow scripting HTTP
 calls if necessary (e.g. query metadata)
 
-Based on [alpine](https://hub.docker.com/_/alpine/) image.
-
 ## Build locally
 
 ```
--- a/bosh-cli/README.md
+++ b/bosh-cli/README.md
@@ -4,8 +4,6 @@ Container for running Bosh SSH.
 * `BOSH` CLI
 * `SSH`
 
-Based on the [ruby:slim](https://hub.docker.com/_/ruby/) image.
-
 ## Build locally
 
 ```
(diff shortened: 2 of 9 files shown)


mtekel:
I remember once updating README to indicate that some image was now based on alpine... And I wondered then already, why need for that if it's in the Dockerfile...

PR #89, opened 2017-02-10
[#121934253] cf-cli: Add blue-green-deploy plugin

dcarley:
## What

We want to use this plugin within our alphagov/paas-cf pipeline. At the
moment it will only be for the healthcheck app, which we never want to have
any downtime because we use it for monitoring and automated tests. We may
choose to use it for other apps in the future if this proves successful.

## How to review

- read the individual commit messages for more information about the changes
- test it with the changes in alphagov/paas-cf#779

## Who can review

Not @dcarley

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -1,8 +1,9 @@
 FROM alpine:3.3
 
 ENV PACKAGES "unzip curl openssl ca-certificates git ruby ruby-json"
+ENV CF_CLI_VERSION "6.23.1"
 
 RUN apk add --update $PACKAGES && rm -rf /var/cache/apk/*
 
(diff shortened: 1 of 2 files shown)


alext:
Is there a test we could add to verify that the plugin is installed correctly?

dcarley:
> Is there a test we could add to verify that the plugin is installed correctly?

`cf blue-green-deploy --help` will return a zero exit code. Do you think that's useful/sufficient?

alext:
> cf blue-green-deploy --help will return a zero exit code. Do you think that's useful/sufficient?

That looks good to me. It verifies that the plugin is installed, and the `cf` can find it etc.

dcarley:
> That looks good to me. It verifies that the plugin is installed, and the cf can find it etc.

👍  added. Shame it doesn't provide a `--version` flag.

PR #90, opened 2017-02-14
bump spruce to version 1.8.9

o-orand:
Could you please consider upgrading spruce to recent version. We use this feature https://github.com/geofffranks/spruce/issues/175, introduced in [1.8.2](https://github.com/geofffranks/spruce/releases/tag/v1.8.2)

Thanks

### DIFF
--- a/spruce/Dockerfile
+++ b/spruce/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.4
 
-ENV SPRUCE_VERSION 1.5.0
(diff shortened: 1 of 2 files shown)


combor:
Hi @o-orand

Thanks for the contribution! I tested this on our cf deploy pipelines and it works without problems so it LGTM.  Merging.

o-orand:
Great.
Thanks @combor !

PR #96, opened 2017-07-17
Compose integration tests

LeePorte:
## What

This adds a version of cf-acceptance-tests that uses Go 1.8.3, this is needed for some of the tests in paas-compose-broker

## How to test
1. Clone the repo
1. Switch to this branch
1. cd cf-acceptance-tests-1.8
1. docker build -t cf-acceptance-tests-1.8 .
1. docker run -i -t cf-acceptance-tests-1.8 go version and confirm it matches go1.8.3 `go version go1.8.3 linux/amd64`

## Who can review

Not me

### DIFF
--- a/cf-acceptance-tests/Dockerfile
+++ b/cf-acceptance-tests/Dockerfile
@@ -1,4 +1,5 @@
 FROM ubuntu:trusty
+ENV DEBIAN_FRONTEND noninteractive
 
 RUN \
   apt-get update && \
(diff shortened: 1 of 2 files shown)


alext:
Why are we using the `cf-acceptance-tests` container to run the Compose integration tests? This feels like it couples the 2 together.

Out of interest, what are the requirements for the Compose tests? Could they not use a vanilla Go container?

Have we tested whether the cf-acceptance-tests work correctly with this updated Go version?

alext:
@LeePorte Is this still needed or can this PR be closed?

LeePorte:
Not needed any longer

PR #101, opened 2017-11-10
[#150711834] Add a new bosh-shell Docker image

henrytk:
## What

https://www.pivotaltracker.com/story/show/150711834

Add a new bosh-shell Docker image

This is intended for use as an interactive shell by an operator on their local machine. This differs from the bosh-cli, which is intended for use in our pipelines.

## How to review

Review in conjunction with https://github.com/alphagov/paas-bootstrap/pull/100
If you don't you will not have the script for testing the container and your Bosh won't have the intended security groups.

* Make sure the latest build of this image in Docker Hub is completed: 
  https://hub.docker.com/r/governmentpaas/bosh-shell/builds/
  It should have a tag corresponding to this branch name (direct-ssh-bosh).
* Use `DEPLOY_ENV=dev make bosh-cli` on the [dev branch of paas-bootstrap](https://github.com/alphagov/paas-bootstrap/tree/direct-ssh-bosh) to run the container.
* Check you can perform all Bosh operations:
  * Basic commands: `bosh vms`
  * Commands against a VM: `bosh do-something vm/index`
  * SSH: `bosh-ssh vm/index`. **Note:** we have added a `bosh-ssh` wrapper script to avoid you having to pass in gateway VM details (to SSH via the Bosh VM).
* Code review
* Travis should take care of the tests. You can refer to the README in the root directory for running them yourself locally.
* Check the way we handle secrets doesn't expose anything it shouldn't.

## Who

Anyone but me or @camelpunch


bosh-shell/startup.sh
```diff
@@ -0,0 +1,26 @@
...
+set -eu
+
+if [ -z "${BOSH_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ]; then
+  echo "WARNING: You need to set BOSH_ID_RSA and BOSH_IP. You will not be logged into Bosh"
+  exec bash
+fi
+
+echo "$BOSH_ID_RSA" | base64 -d > /tmp/bosh_id_rsa && chmod 400 /tmp/bosh_id_rsa
+
+ssh -fNT -L 25555:localhost:25555 \
```
alext: I'm seeing a warning `bind: Address not available` from this command. It seems that it's because it tried to listen on IPv6 localhost as well as IPv4, but this isn't available within the container. I can silence this warning by using the `-4` option to ssh.

bosh-shell/bosh-ssh
```diff
@@ -0,0 +1,17 @@
...
+    exit 1
+fi
+
+target=$1
+
+exec bosh ssh \
+   "$target" \
+   --gateway_host "$BOSH_IP" \
+   --gateway_user vcap \
+   --gateway_identity_file /tmp/bosh_id_rsa
```
alext: When running this command, I see lots of deprecation warnings output:
```
/usr/local/bundle/gems/net-ssh-2.9.2/lib/net/ssh/transport/session.rb:67:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
/usr/local/bundle/gems/net-ssh-2.9.2/lib/net/ssh/transport/session.rb:84:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
/usr/local/bundle/gems/net-ssh-2.9.2/lib/net/ssh/transport/cipher_factory.rb:98: warning: constant OpenSSL::Cipher::Cipher is deprecated
/usr/local/bundle/gems/net-ssh-2.9.2/lib/net/ssh/transport/cipher_factory.rb:72: warning: constant OpenSSL::Cipher::Cipher is deprecated
```

I suspect that this is because the version of bosh-cli that we're using was developed against an earlier version of ruby.  I wonder whether we should use ruby 2.2 for this container, which would match the version used in the bosh-cli container.

camelpunch: We made a conscious decision to keep these Ruby warnings as positive pressure to upgrade to modern bosh-cli. Ruby bosh-cli isn't maintained, and [soon Ruby 2.2 won't be either](https://www.ruby-lang.org/en/downloads/).

bosh-shell/startup.sh
camelpunch: Good catch! We were wondering where that came from.

bosh-shell/bosh-ssh
alext: I'm not convinced that's the right choice in this case. Those warnings are a lot of noise, and make it harder to see what's going on when using this.

Secondly, I think it's better that this container is consistent with the one used in the pipeline. At some point we'll have to update all our containers to use a newer ruby, but I think that's better done in one go so that we don't have a mixture of versions in use.

bosh-shell/startup.sh
henrytk: Nice one. Fixed.

bosh-shell/bosh-ssh
henrytk: Reverted back to 2.2

PR #103, opened 2017-12-14
[#153091168] Add autopilot plugin to cf-cli docker image

46bit:
## What

We are updating the product page deployment to deploy with a host of `www.cloud.service.gov.uk` (see https://github.com/alphagov/paas-product-page/pull/44.) For this we need the [autopilot](https://github.com/contraband/autopilot) `cf` plugin to be available in `cf-cli` Docker image [being used](https://github.com/alphagov/paas-release-ci/blob/master/pipelines/deploy-app.yml#L91-L98).

## How to review

Code review

## Who can review

Not @46bit @camelpunch

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -5,6 +5,9 @@ ENV CF_CLI_VERSION "6.26.0"
 ENV CF_BGD_VERSION "1.1.0"
 ENV CF_BGD_CHECKSUM "fe6ebd3c2dc3a287db0d31eeaed200d1a27117d9a6ddd875de561e4a6e8858d1"
 ENV CF_BGD_TEMPFILE "/tmp/blue-green-deploy.linux64"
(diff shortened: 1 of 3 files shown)


46bit:
@bandesz Sorry, that context was missed. The `blue-green-deploy` plugin ignores new settings in the app manifest. We have to add new routes to the `manifest.json` and have them be applied. Autopilot only pays attention to the app manifest, so we've switched to use it here.

bandesz:
I'm going to approve this now, but just a note that this plugin is not maintained at all and Trade Tariff  also reported a strange error: https://github.com/contraband/autopilot/issues/49. We could write the same deployment process with a couple of cf commands.

camelpunch:
There's now a version 0.0.4 merged into the official repo: https://github.com/cloudfoundry/cli-plugin-repo/pull/193

PR #107, opened 2018-01-09
[#153119879] Upgrade Terraform to latest stable

dcarley:
## What

Upgrade Terraform to latest stable. See individual commits for more information.

## How to review

1. Review the commits individually.
1. Test with other PRs on paas-cf, paas-bootstrap, and paas-release-ci.

## Who can review

Not @dcarley or @46bit

### DIFF
--- a/terraform/plugin_cache.tf
+++ b/terraform/plugin_cache.tf
@@ -1,0 +1,12 @@
+provider "aws" {
+  version = "1.6.0"
+}
+provider "datadog" {
+  version = "1.0.3"
+}
+provider "template" {
+  version = "1.0.0"
+}
+provider "null" {
+  version = "1.0.0"
+}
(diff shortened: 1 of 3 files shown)


46bit:
👍 I'd like to pin the versions. What do you think, @dcarley?

dcarley:
Yep 👍  Are you OK to make the change, @46bit?

46bit:
@samcrang The container built with the versions 👍

terraform/plugin_cache.tf
```diff
@@ -0,0 +1,12 @@
+provider "aws" {
+  version = "~> 1.6"
```
samcrang: does it make sense to just use [exact version equality](https://www.terraform.io/docs/configuration/terraform.html#specifying-a-required-terraform-version)?

46bit: Sure. I've pinned this to the versions that the DockerHub build chose https://hub.docker.com/r/governmentpaas/terraform/builds/bxwtfu2lbnwwqdjt8ktmynw/ :-)

PR #108, opened 2018-01-11
bump spruce to 1.14 and alpine to 3.7

o-orand:
Using the latest spruce version, doesn't require usage of ! to escape  BOSH/Concourse
variables.


samcrang:
Looks good, thanks! 👍

o-orand:
Thanks 😄

PR #109, opened 2018-01-11
bosh-shell improvements

dcarley:
### bosh-shell: Automatically target BOSH deployment

So that you don't have to run these commands yourself each time you use
`make bosh-cli` from paas-bootstrap, which I find very frustrating. The old
target in paas-cf used to do an equivalent of this for you.

This requires a new environment variable to be passed from paas-bootstrap.
I've called it something other than `DEPLOY_ENV` to make it easier to
understand what it is used for.

### bosh-shell: Use a wrapper script for bosh ssh

Rather than requiring people to run a separate command because I keep
forgetting this. This works because `exec bosh` within the script doesn't
use the `alias` set in the shell.

The existing `bosh --version` spec test should make sure that normal bosh
usage still works. I don't think we can do much more without a director.

## How to review

See alphagov/paas-bootstrap#118

## Who can review

Anyone, but possibly @alext while on support.

### DIFF
--- a/bosh-shell/Dockerfile
+++ b/bosh-shell/Dockerfile
@@ -4,7 +4,9 @@ RUN gem install bosh_cli -v 1.3262.26.0 --no-rdoc --no-ri
 
 RUN apk add --update bash openssl openssh-client file git && rm -rf /var/cache/apk/*
 
-COPY startup.sh /startup.sh
 COPY bosh-ssh /usr/local/bin
+COPY bosh_wrapper.sh /usr/local/bin
+COPY bosh_wrapper.rc /
+COPY startup.sh /startup.sh
 
 ENTRYPOINT /startup.sh
--- a/bosh-shell/bosh-ssh
+++ b/bosh-shell/bosh-ssh
@@ -1,17 +1,4 @@
 #!/bin/bash
 
-set -eu
-
-if [ $# -eq 0 ]
-then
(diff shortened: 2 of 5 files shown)


bosh-shell/Dockerfile
```diff
@@ -4,7 +4,8 @@ RUN gem install bosh_cli -v 1.3262.26.0 --no-rdoc --no-ri
 
 RUN apk add --update bash openssl openssh-client file git && rm -rf /var/cache/apk/*
 
+COPY bosh_wrapper.sh /usr/local/bin
+COPY bosh_wrapper.rc /
 COPY startup.sh /startup.sh
-COPY bosh-ssh /usr/local/bin
```
alext: Is it worth leaving a command called bosh-ssh that outputs a helpful message telling people that it's no longer necessary (could even be an alias)

alext:
I've run into issues (unrelated to these changes) with the ssh connection being dropped after being idle for 60 seconds. I've added a commit on the `bosh_shell_improvements_alext` branch to reduce the `ServerAliveInterval` to 30, which seems to improve things. What do you think about including that in this PR?

bosh-shell/Dockerfile
dcarley: Good idea, done. I used a command because it was easier to set the correct return code.

dcarley:
> What do you think about including that in this PR?

👍  I've cherry-picked

PR #110, opened 2018-01-25
Update psql to Alpine 3.4

camflan:
updating to Alpine 3.4 so k8s dns will work

### DIFF
--- a/psql/Dockerfile
+++ b/psql/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.3
+FROM alpine:3.4
 
 ENV PACKAGES "postgresql-client"
 


samcrang:
Hi @camflan 

I think this looks like a reasonable thing to do—would you be able to expand a bit more on what's broken with Alpine <3.4 and k8s DNS though?

The tests are failing because we assert on the Alpine version but this should be an easy fix.

Sam

camflan:
@samcrang 3.4+ supports search domains in `resolv.conf`, which allows the usage of kube-dns in a cluster. Bumping to 3.6 might be good too, if you're going to bump the versions anyway.

thanks!

samcrang:
@camflan I've created a new PR (#111) that includes this change, takes things to the latest version of Alpine and fixes the tests. I'm going to close this one off.

Thanks for the PR!

PR #113, opened 2018-02-06
bosh-shell: handle non-existing CF deployment

alext:
## What

Currently the startup script attempts to download the bosh manifest for
the corresponding CF deployment and target it. If this deployment
doesn't exist, it causes the download to fail, and the startup script to
abort. As a result it's not possible to use this when the CF deployment
doesn't exist.

This updates the script to gracefully handle a missing CF deployment so
that you can still interact with bosh.

## How to review

Review along with paas-bootstrap PR TBC

## Who can review

Not me.

### DIFF
--- a/bosh-shell/startup.sh
+++ b/bosh-shell/startup.sh
@@ -19,7 +19,11 @@ ssh -fNT -4 -L 25555:localhost:25555 \
 bosh -t localhost login admin -- "$BOSH_ADMIN_PASSWORD"
 bosh target localhost
 
-bosh download manifest "${BOSH_DEPLOYMENT}" "${BOSH_DEPLOYMENT}.yml"
-bosh deployment "${BOSH_DEPLOYMENT}.yml"
+if bosh download manifest "${BOSH_DEPLOYMENT}" "${BOSH_DEPLOYMENT}.yml"; then
+  bosh deployment "${BOSH_DEPLOYMENT}.yml"
+else
+  echo "No deployment targetted..."
+  echo "You'll need to manually download a manifest and target a deployment."
+fi
 
 exec bash --rcfile /bosh_wrapper.rc


henrytk:
When I run this (with no CloudFoundry deployment and a failing Concourse deployment) I get the following error:

```
Warning: Permanently added 'redacted IP' (ECDSA) to the list of known hosts.
Unauthorized use is strictly prohibited. All access and activity
is subject to logging and monitoring.
Logged in as 'admin'
Target set to 'henry'
Acting as user 'admin' on deployment 'henry' on 'henry'
Deployment manifest saved to 'henry.yml'
Incorrect YAML structure in '/henry.yml': expected Hash at the root
```

bosh-shell/startup.sh
```diff
@@ -19,7 +19,11 @@ ssh -fNT -4 -L 25555:localhost:25555 \
 bosh -t localhost login admin -- "$BOSH_ADMIN_PASSWORD"
 bosh target localhost
 
-bosh download manifest "${BOSH_DEPLOYMENT}" "${BOSH_DEPLOYMENT}.yml"
-bosh deployment "${BOSH_DEPLOYMENT}.yml"
+if bosh download manifest "${BOSH_DEPLOYMENT}" "${BOSH_DEPLOYMENT}.yml"; then
+  bosh deployment "${BOSH_DEPLOYMENT}.yml"
```
henrytk: Suggested fix if this is an empty file:

```
if bosh download manifest "${BOSH_DEPLOYMENT}" "${BOSH_DEPLOYMENT}.yml"; then
  if [ -s "${BOSH_DEPLOYMENT}.yml" ]; then
    bosh deployment "${BOSH_DEPLOYMENT}.yml"
  else
    echo "Unable to target deployment ${BOSH_DEPLOYMENT}, manifest is an empty file."
  fi
else
  echo "No deployment targetted..."
  echo "You'll need to manually download a manifest and target a deployment."
fi
```

alext: Good call. I've added a commit to with this fix.

PR #122, opened 2018-04-05
[Firebreak] Add image for executing arbitrary Bosh commands

henrytk:
This allows you to execute one-off commands using Bosh. The initial
motivation for this comes from the repository paas-infrastructure-map,
which provides the tools necessary to query Bosh and build a dynamic
map of a Bosh deployment.

We use the empty exec-style `CMD` declaration to override the command
of the base Ruby image.

### DIFF
--- a/bosh/Dockerfile
+++ b/bosh/Dockerfile
@@ -1,0 +1,18 @@
+FROM ruby:2.2-alpine
+
+ENV BOSH_CLI_VERSION 3.0.1
+ENV BOSH_CLI_SUM ccc893bab8b219e9e4a628ed044ebca6c6de9ca0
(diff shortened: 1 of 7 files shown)


bosh/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ruby:2.2-alpine
+
+ENV BOSH_CLI_VERSION 2.0.48
```
LeePorte: Can we use Bosh CLI 3.0.1 due to https://www.cloudfoundry.org/blog/cve-2018-1231/

henrytk: Job's a good'n

henrytk:
I've also cleaned up the bosh-shell README because it is out of date.

PR #124, opened 2018-04-13
[#155037639] Add spruce to cf-cli for manifest templating

keymon:
What?
----

I want to generate the manifest of the product page, pulling  secrets from .yml files in S3 buckets as we still don't have any sensible proper solution for this.

spruce[1] is a nice tool to kickly generate manifests, pull secrets from other variables, etc. It is handy to have it in the cf-cli container to simplify our pipelines, so that we can build a manifest and add the required config (env vars, secrets, routes, etc) seamlessly.

We also want to add bash, as it has more capabilities than the simple `sh`, we can write simpler scripts. The additional consume sized is minimal.

[1] https://github.com/geofffranks/spruce

How to review?
--------------

Code review

Who?
---

Anyone but me

### DIFF
--- a/cf-cli/Dockerfile
+++ b/cf-cli/Dockerfile
@@ -1,6 +1,6 @@
 FROM alpine:3.3
 
-ENV PACKAGES "unzip curl openssl ca-certificates git ruby ruby-json libc6-compat"
+ENV PACKAGES "unzip curl openssl ca-certificates git ruby ruby-json libc6-compat bash"
 ENV CF_CLI_VERSION "6.34.1"
 ENV CF_BGD_VERSION "1.3.0"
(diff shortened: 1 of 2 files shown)


cf-cli/Dockerfile
```diff
@@ -28,3 +29,10 @@ RUN curl -L -o "${CF_AUTOPILOT_TEMPFILE}" \
   && chmod +x "${CF_AUTOPILOT_TEMPFILE}" \
   && cf install-plugin -f "${CF_AUTOPILOT_TEMPFILE}" \
   && rm "${CF_AUTOPILOT_TEMPFILE}"
+
+RUN apk add --update wget ca-certificates \
```
bandesz: ca-certificates is already installed and probably you should not remove it at the end

bandesz: also curl is available so, use that instead of wget

keymon:
comments fixed

keymon:
> see my comments

Fixed

PR #134, opened 2018-09-19
[#160614327] Update bosh-cli-v2 to 5.2.1 with SKI for certs

keymon:
What?
-----

The version 5.2.1 [1] supports generating certificates via `--vars-store` with SKI/AKI.

We need this to be able to rotate CA certificates used by CF API, otherwise
it is not possible to use two CA certs with the same name at the same time.

More info in [2]

[1] https://github.com/cloudfoundry/bosh-cli/releases/tag/v5.2.1
[2] https://github.com/cloudfoundry/config-server/pull/9

The upstream project for bosh-cli now

How to review
-------------

Check that the new container generate certs with the SKI. You can do so
with this script:

```
test_bosh_cli_cert() {
    image=$1
    manifest='
cert: ((test.certificate))

variables:
- name: test
  options:
    common_name: foo
    is_ca: true
  type: certificate
'
    echo "${manifest}" | \
        docker run -i ${image} bosh interpolate - --vars-store=/ignore.yml | \
        ruby -ryaml -e 'puts YAML.load(STDIN)["cert"] ' | \
        openssl x509 -text -noout -in - | \
        grep -A 4 "Subject Key Identifier"

}

test_bosh_cli_cert governmentpaas/bosh-cli-v2:latest

test_bosh_cli_cert governmentpaas/bosh-cli-v2:update_bosh_cli-160614327

```

What next?
---------

Once merged, a new container id would be gerated and paas-cf needs updating.

Who?
----

Not me

### DIFF
--- a/bosh-cli-v2/Dockerfile
+++ b/bosh-cli-v2/Dockerfile
@@ -1,7 +1,7 @@
 FROM ruby:2.5-slim
 
-ENV BOSH_CLI_VERSION 3.0.1-gds1
-ENV BOSH_CLI_SUM fb71095c50aaf60865a4687cf22f2c826f44ed2f
+ENV BOSH_CLI_VERSION 3.0.1
+ENV BOSH_CLI_SUM ccc893bab8b219e9e4a628ed044ebca6c6de9ca0
 ENV BOSH_CLI_FILENAME bosh-cli-${BOSH_CLI_VERSION}-linux-amd64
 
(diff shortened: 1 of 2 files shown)


keymon:
waiting for https://hub.docker.com/r/governmentpaas/bosh-cli-v2/builds/bdqhaxoavpkqqlgq2ivyg5k/

PR #135, opened 2018-09-20
[#160614327] Update bosh-cli for bosh and bosh-shell

keymon:
What?
----

We have update the bosh-cli-v2 to use bosh-cli 5.2.1 with SKI for
certs, and we want to use the same in the rest of containers
for consistency.

The version 5.2.1 [1] supports generating certificates via
--vars-store with SKI/AKI.

We need this to be able to rotate CA certificates used by CF API,
otherwise it is not possible to use two CA certs with the same name
at the same time.

More info in [2]

[1] https://github.com/cloudfoundry/bosh-cli/releases/tag/v5.2.1
[2] https://github.com/cloudfoundry/config-server/pull/9

How to review?
-----

code review, tests pass

Who?
----

not me

### DIFF
--- a/bosh-shell/Dockerfile
+++ b/bosh-shell/Dockerfile
@@ -1,7 +1,7 @@
 FROM ruby:2.5-alpine
 
-ENV BOSH_CLI_VERSION 2.0.48
-ENV BOSH_CLI_SUM c807f1938494f4280d65ebbdc863eda3f883d72e
+ENV BOSH_CLI_VERSION 5.2.1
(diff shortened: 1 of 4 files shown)


keymon:
Waiting for:

https://hub.docker.com/r/governmentpaas/bosh/builds/bokxcqtlwrrnchlrraetlnh/
https://hub.docker.com/r/governmentpaas/bosh-shell/builds/b8swi57jagd8vvxrxcx2etd/

PR #142, opened 2018-11-13
[#161544435] Upgrade cf-log-cache plugin and cf-cli

keymon:
What?
----

During the upgrade of cf-deployment to v6.0.0[1] log-cache has been
upgraded. The new version of log-cache uses a new protocol
and it is not compatible with old clients.

We need to update the log-cache cf client plugin that runs on the
tests.

[1] https://github.com/cloudfoundry/cf-deployment/releases/tag/v6.0.0

How to review
-----------

Code review. Travis should pass

Who?
----

Not me


keymon:
That worked

PR #147, opened 2019-01-02
Increment the version of CF_CLI

tnwhitwell:
# What
Increment the version of cf-cli in cf-cli docker image to allow use of `--assign-stack` argument.

# How to review
* Code Review
* Test by building the image and using cf from it

# Who can review
Not me


keymon:
LGTM

PR #148, opened 2019-01-25
[#163263881] Use custom AWS CPI version

bandesz:
## What

We want to use the r5d instance types, but the AWS CPI does not have support for it.

## How to review

TODO

## Who can review it

Not me.


alext:
Closing because we need support in upstream before we can use the r5d instance types

PR #149, opened 2019-04-09
AUTO: Build cf cli from source

andy-paine:
- There are some commits that would be handy for Notify on latest of the
  cf cli but they haven't release binaries yet
- Use a multistage build to build the cli from source and then copy it
  into the old image rather than curl-ing it


tlwr:
Closing this as we are using 6.45 now and because the tests aren't passing. :D

PR #150, opened 2019-04-09
BAU: Update to use bundler 2

andy-paine:
- The push/PR builds are failing due to bundler not being installed
- Travis docs here
  https://docs.travis-ci.com/user/languages/ruby/#bundler-20 suggest
that we need to install bundler 2.0 explicitly


tlwr:
`bundle update --bundler` also required?

PR #152, opened 2019-04-11
BAU: Fix broken Travis build

andy-paine:
- Uses @richardTowers Ruby pinning as that appears to be important
- Actually fixes broken test too (see commit message for what was going wrong)
- I've upgraded the awscli and changed the test as it was lacking loads of features but happy for that to be done separately if that is a contentious change

Closes #150
Closes #151

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - 2.5
+  - 2.5.1
 
 sudo: required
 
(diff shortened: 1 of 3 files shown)


richardTowers:
Fixes https://www.pivotaltracker.com/n/projects/1275640

tlwr:
I think you need to be more specific with the link

PR #165, opened 2019-06-19
Upgrade bosh-cli in bosh-cli-v2 and bosh-shell

tlwr:
We would like the latest version of bosh-cli in our bosh container images.


tlwr:
Merging this as it doesn't actually get it deployed anywhere and all the tests pass

PR #170, opened 2019-09-18
[WIP] Create dedicated credhub-shell container

AP-Hunt:
Credhub and the Bosh director are co-located, but in some instances we don't
want to use a container which has access to the Bosh CLI just to use the
Credhub CLI. This container addresses that need.


AP-Hunt:
Closing. Decided to take a different path.

PR #173, opened 2019-10-01
[Firebreak] Add jq to curl-ssl image

tnwhitwell:
It would be really useful to have `jq` available in the `curl-ssl` image, so json responses can be parsed easily.

This change adds that tool.


richardTowers:
Could you update the "spec" too? 

https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/master/curl-ssl/curl-ssl_spec.rb

PR #175, opened 2019-10-09
Update cf cli to v 6.46.1

o-orand:
Bump version of cf-cli used in governmentpaas/cf-cli container to
latest (currently v6.46.1, v7.x still in beta)


tlwr:
Applied your commit as a patch locally, and the tests pass.

Thanks!

PR #178, opened 2019-11-20
[#166500017] Add option to pass in a user's key

mogds:
What
---

We're switching to use individual accounts instead of shared creds. This updates the bosh-shell image with a USER_ID_RSA var so we can pass in user specific keys.

How to review
---

- Code review
- Build locally and update image used [here](https://github.com/alphagov/paas-bootstrap/compare/indiviual-accounts-166500017?expand=1#diff-1f5c6bd196a437e4dca5a2c70344b48cL50)

Who can review
---

Not me

### DIFF
--- a/bosh-shell/startup.sh
+++ b/bosh-shell/startup.sh
@@ -2,13 +2,13 @@
 
 set -eu
 
-if [ -z "${BOSH_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
-  echo "WARNING: You need to set BOSH_ID_RSA, BOSH_IP, BOSH_ENVIRONMENT and BOSH_CA_CERT."
+if [ -z "${USER_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
+  echo "WARNING: You need to set USER_ID_RSA, BOSH_IP, BOSH_ENVIRONMENT and BOSH_CA_CERT."
   echo "You will not be logged into Bosh."
   exec bash
 fi
 
-echo "$BOSH_ID_RSA" | base64 -d > /tmp/bosh_id_rsa && chmod 400 /tmp/bosh_id_rsa
+echo "$USER_ID_RSA" | base64 -d > /tmp/user_id_rsa && chmod 400 /tmp/user_id_rsa
 
 ssh -fNC -4 -D 25555 \
   -o ExitOnForwardFailure=yes \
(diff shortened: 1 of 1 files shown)


bosh-shell/startup.sh
```diff
@@ -2,25 +2,26 @@
 
 set -eu
 
-if [ -z "${BOSH_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
-  echo "WARNING: You need to set BOSH_ID_RSA, BOSH_IP, BOSH_ENVIRONMENT and BOSH_CA_CERT."
+if [ -z "${USER_ID_RSA:-}" ] || [ -z "${BOSH_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
```
richardTowers: Do we still need BOSH_ID_RSA as well here? Looks like it's only going to use the USER_ID_RSA one.

mogds: Was planning on removing it in a follow up PR after a paas-cf deployment

tlwr: We reference the hash explicitly, so breaking changes are fine :)

bosh-shell/startup.sh
```diff
@@ -2,25 +2,25 @@
...
-if [ -z "${BOSH_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
-  echo "WARNING: You need to set BOSH_ID_RSA, BOSH_IP, BOSH_ENVIRONMENT and BOSH_CA_CERT."
+if [ -z "${USER_ID_RSA:-}" ] || [ -z "${BOSH_IP:-}" ] || [ -z "${BOSH_ENVIRONMENT:-}" ] || [ -z "${BOSH_CA_CERT:-}" ]; then
+  echo "WARNING: You need to set USER_ID_RSA, BOSH_IP, BOSH_ENVIRONMENT and BOSH_CA_CERT."
   echo "You will not be logged into Bosh."
   exec bash
 fi
 
-echo "$BOSH_ID_RSA" | base64 -d > /tmp/bosh_id_rsa && chmod 400 /tmp/bosh_id_rsa
+echo "$USER_ID_RSA" | base64 -d > /tmp/user_id_rsa && chmod 400 /tmp/user_id_rsa
```
AP-Hunt: Why do we base64 this?

tlwr: Probably `docker run -e USER_ID_RSA=<base64 encoded>`

AP-Hunt: Makes sense. Tested and it works.

PR #182, opened 2020-01-15
bump to latest awscli version

philandstuff:
My selfish reason for wanting this is to get support for the [k8s IAM
magical auth support][1] but it's probably good to keep updating this
for all sorts of reasons.

1.17.0 dropped support for python 2.6 and 3.3; I don't think that's a
concern for us here?

[1]: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html


tlwr:
LGTM :shipit:

PR #187, opened 2020-06-04
Update alpine to 3.12

tlwr:
3.9 is old 3.12 is latest as of 29-05-2020


tlwr:
Tests pass, merging this

PR #200, opened 2020-10-29
Upgrade terraform to 0.13.3

richardTowers:
The GOV.UK replatforming team are using 0.13.3, but we're too lazy to
build our own docker images at this point in time.


LeePorte:
superceeded by https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/206

PR #202, opened 2020-11-24
Bump awscli version

Krenair:
Per https://github.com/boto/boto3/issues/2596#issuecomment-694508466

The PaaS might not use this but we've seen it when re-using the image for
OctoDNS and trying to `pip3 install boto3` on top of governmentpaas/awscli.


46bit (approved):
PaaS don't rely on anything special about the `aws-cli` version so we're fine to adopt this.

PR #205, opened 2020-12-16
bosh-cli-v2-cf-cli: update to cf to 7.2.0; spruce to 1.27.0

o-orand:
bump to latest available cli. We also include autoscaler plugin.


o-orand:
Sorry wrong PR

PR #206, opened 2020-12-18
Update terraform to last version in the 0.13 series

LeePorte:
Terraform 0.14 is out, so we should start upgrading


AP-Hunt (approved):
Checksum matches the one published for `terraform_0.13.5_linux_amd64.zip`

PR #207, opened 2020-12-22
[#176226306] Build mirrored DockerHub images for us to use in production

46bit:
## What

DockerHub recently started rate limiting. As a result we are stopping using DockerHub for high-volume production CI/CD pipelines. This commit finishes off that work.

This commit creates 3 new Dockerfiles corresponding to the 3 remaining uses of DockerHub in our pipelines. We want to use GitHub Container Registry, but these external images aren't available there yet.

Alternative is to have something copy these 3 upstream images across to GitHub Container Registry, but that doesn't fit with the existing pattern we have in this repo.

## How to review

See if things go green, and decide if you're OK with this approach.

Once this is merged there's then https://github.com/alphagov/paas-release-ci/pull/162 to review, which builds these images and pushes them to GCHR.

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -1,0 +1,1 @@
+FROM alpine:3.12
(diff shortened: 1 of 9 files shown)


46bit:
I've discovered that the Concourse pipeline we use to build containers for GitHub Container Registry is static, not automatically generated. So I'll edit that to copy the images across rather than doing this.

PR #203, opened 2020-12-14
cf-cli: update to v7.2.0; spruce to v1.27.0

o-orand:
Hi There!
A PR to bump to latest cli available (cf and spruce)

PR #208, opened 2020-12-22
spruce: bump to v1.27.0

o-orand:
Spruce bump is required to be consistent with embedded spruce in cf-cli image.
This is related to  #203


LeePorte (approved):
Thanks for the update, and sorry about the delay in getting to it.

o-orand:
No problem, I forgot in first place to include this change everywhere.

PR #217, opened 2021-07-13
Add node-chromium image

paroxp:
Essentially we need a headless browser to be performing some of our acceptance tests in.

### DIFF
--- a/node-chromium/Dockerfile
+++ b/node-chromium/Dockerfile
@@ -1,0 +1,12 @@
+FROM node:lts-alpine
+
+RUN apk add --no-cache \
+    chromium \
+    nss \
+    freetype \
+    harfbuzz \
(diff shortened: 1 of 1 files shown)


kr8n3r (approved):

node-chromium/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:lts-alpine
+
+RUN apk add --no-cache \
```
0atman, line 3: Sorry I missed this pre-merge.

Should we && `apk update && apk upgrade` before this?

kr8n3r: was subsequently added https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/218/files

PR #222, opened 2022-03-21
Update cf-acceptance-tests to cf-cli 7.4

whpearson:
We found that double logouts were causing test suite failures for the
autoscaler-acceptance-tests due to the cli version. Version 7.4.0 is
found not to have that issue.

The location for the debian package was retrieved from
https://github.com/cloudfoundry/cli/releases as it later versions were
not being found at the previous location.


risicle (approved):
Huzzah!

PR #223, opened 2022-05-24
Upgrade terraform to 1.2.0

LeePorte:
We like to keep current with our terraform, this PR upgraded both terraform and the modules used


terraform/plugin_cache.tf
```diff
@@ -2,14 +2,14 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "2.70.0"
+      version = "4.15.0"
```
AP-Hunt, line 5: Are we confident there are no breaking changes between 2.x and 4.x that will affect us?

LeePorte: Based on what I've seen in aws-account-wide there should be no issues. As the containers are referenced by hash in the pipelines, the plan was to test it once built in dev and then upgrading any of our terraform in paas-cf and paas-bootstrap as required.

AP-Hunt (approved):

PR #224, opened 2022-05-31
Build images with GitHub actions

whi-tw:
Rather than building these on concourse, instead build them with github actions.

This PR currently only pushes the images to ghcr.io. Could be extended to also push to docker hub, although from what I can see a lot of projects are moving away from dockerhub.

Additionally, build a version of cf-acceptance-tests with ginkgo v2 as well as ginkgo v1

Action runs on my fork are visible here: https://github.com/whi-tw/paas-docker-cloudfoundry-tools/actions?query=event%3Apush
Images built on my fork are visible here: https://github.com/whi-tw?tab=packages&repo_name=paas-docker-cloudfoundry-tools

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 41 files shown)


AP-Hunt (approved):
I think this looks really neat. We ought to have a second PR to retire the Concourse pipelines currently doing the same job, and fix up any docs we have that point to that.

whi-tw:
> I think this looks really neat. We ought to have a second PR to retire the Concourse pipelines currently doing the same job, and fix up any docs we have that point to that.

Also, we will need to point the pipelines / tooling to use the images from ghcr.io, not dockerhub.

PR #232, opened 2022-06-09
Enable linting on changed files

whi-tw:
Run superlinter over all changes files in commits / PRs.

This should catch any silly flubs with actions, as well as any other big errors.


AP-Hunt (approved):
I've not heard of super-linter before, but I get the point. Should we consider it for other repos?

PR #233, opened 2022-06-09
[#182295946] Upgrade alpine to 3.16

whi-tw:
## What

Upgrade alpine to 3.16 across containers which use it as a base image.

## How to review

* Code review
* Check tests have run

## Before Merging

* This should be merged before #236 
* the "REMOVE_ME_PRE_MERGE" commit needs to be dropped before merging

## After Merging

* Move on to #237

### DIFF
--- a/.github/workflows/image_alpine.yml
+++ b/.github/workflows/image_alpine.yml
@@ -25,4 +25,5 @@ jobs:
     uses: ./.github/workflows/build-image-base.yml
     with:
       image: alpine
+      platforms: linux/amd64,linux/arm64
     secrets: inherit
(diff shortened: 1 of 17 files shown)


LeePorte:
We should build and push on PRs for the purpose of being able to use the containers in a pipeline to validate that they work before merging. However they shouldn't be tagged as latest

whi-tw:
> We should build and push on PRs for the purpose of being able to use the containers in a pipeline to validate that they work before merging. However they shouldn't be tagged as latest

I've implemented these changes in #234 - do you mind giving it a look?

whi-tw:
> @whi-tw All of the jobs have built and pushed containers now. I'm dropping the temp commit and force pushing.

As-per Slack, build is working now. have split curl-ssl to #237, so that's next!

AP-Hunt (approved):

PR #241, opened 2022-07-21
182427769 - Upgrade to Ginkgo V2

nimalank7:
Description:
- Bumps Ginkgo V2 Dockerfile to 1.18
- Removes Ginkgo V1 GitHub action and Dockerfile as no longer necessary
- Removes path workflow as otherwise GitHub actions will only be triggered when a change is made to a particular subdirectory. If that is the case then we will have differing hashes for different subdirectories but we need the same hashes to update paas-cf docker images. So we trigger everything when any change is made to paas-docker-cloudfoundry-tools
- Updates dependencies to Docker images to newer versions as Alpine Linux removes old versions which will fix broken docker builds and removes OpenSSL from the git-ssh image as it is already present in the base alpine image


cadmiumcat (approved):
👍

cadmiumcat:
It'd be helpful if you could split the PR into a few commits (maybe corresponding to how you've described them above).  

I'd usually do this with `git reset HEAD~`, and then adding the individual commits

PR #243, opened 2022-07-28
182427769 - Upgrade OpenSSL alpine base image

nimalank7:
Description:
- Older OpenSSL image isn't present in the Alpine repository so we have to bump up to ensure the GitHub actions pass


cadmiumcat (approved):
I'm approving this, but we have agreed that it will require another PR. We need a fresh version on `main` for alpine before we can move forward

PR #248, opened 2022-10-05
Upgrade cf acceptance test base image to bionic

schmie:
## What
Upgrade cf acceptance test base image to bionic

This is one of the last places xenial was used. Broke our billing acceptance tests as postgres is no longer providing xenial apt packages.

## How to review

- Check CI
- Deploy in dev env


whi-tw:
superceeded by #249 and #250

PR #252, opened 2022-10-06
set apt to be noninteractive

whi-tw:
## What

export `DEBIAN_FRONTEND=noninteractive` in the base image, so that apt behaves as expected in child images

## How to review

- Sense check


nimalank7 (approved):
LGTM!

PR #260, opened 2022-10-28
constrain versions used by `apk` less restrictively

whi-tw:
## What

Previously we had 'hard' version pinning for `apk` installed packages, which means that as soon as those versions fall off the repositories, our builds started failing.

Instead of this, I have changed the pinning so that it will get either the latest version that matches a certain semver tilde restriction (these can be visualised with the [Online Semver checker](https://jubianchi.github.io/semver-check/#/constraint/~7)).

## How to review

Check that the constraints I have added make sense

### DIFF
--- a/alpine/Dockerfile
+++ b/alpine/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:3.16
 
 RUN apk add --no-cache \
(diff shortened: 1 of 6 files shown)


cadmiumcat (approved):
yes, please make the version pain stop

PR #262, opened 2022-11-04
Use Ruby 3.1 across the board

AP-Hunt:
## What

Wherever we use Ruby in a Docker image, upgrades to Ruby 3.1

## How to review

Check the images are built successfully, and tests pass.

N.B. because some images (such as `cf-cli`) reference `ghcr.io/alphagov/paas/ruby-base:main`, they won't _actually_ get Ruby 3.1 until this PR is merged


schmie (approved):
LGTM - CI is happy and pr-262 builds exist in ghcr  👍

PR #264, opened 2022-11-23
[#183708862] Use Ruby 3.1 for tests

AP-Hunt:
## What

As a follow up to #262 and #263, uses Ruby 3.1 for running the tests against the containers. For the sake of consistency with the rest of the PaaS real estate.

## How to review

All the checks passing? Do the build logs suggest Ruby 3.1 is being used?


AP-Hunt:
> It's unclear if any of the workflows actually use ruby - every workflow run here seems to skip the "set up ruby" step.

It's done in the reusable [build-image-base.yml](https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/main/.github/workflows/build-image-base.yml#L120) workflow

risicle (approved):
Ok so it looks like it must be working

PR #285, opened 2023-07-10
185509385 - Pin GitHub Actions to specific hashes

nimalank7:
Description:
- Currently we pin to versions which means that we automatically pull in the latest changes which presents a security risk as we don't know which code is running in our build pipeline.
- This PR fixes this by pinning to a specific hash

## How to review

Verify that GitHub actions successfully run


fearoffish (approved):
lgtm

PR #314, opened 2023-11-06
updating ubuntu bionic->jammy

jackjoy-gds:
## What

we need to update these refs to the old bionic ubuntu versions as the apt repos seem to have been removed, which is preventing the pipelines running on the billing integration tests...

## How to review

Look at the code


corlettb (approved):
We need to go to jammy.

PR #318, opened 2023-11-29
[186049138] Postgresql is used in acceptance tests

dark5un:
## What

We are trying to eliminate any apt-get install remnants in our pipelines by loading requirements in the docker images.
This PR will include postgresql and postgresql-client needed in acceptance-tests.

## How to review

Verify that the image contains postgresql and postgresql-client installed.

### DIFF
--- a/cf-acceptance-tests/cf-acceptance-tests_spec.rb
+++ b/cf-acceptance-tests/cf-acceptance-tests_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper'
 require 'docker'
 require 'serverspec'
+require 'pathname'
 
 GO_VERSION="1.21.1"
 CF_CLI_VERSION="8.6.0"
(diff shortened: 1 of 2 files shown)


cf-acceptance-tests/cf-acceptance-tests_spec.rb
```diff
@@ -58,6 +59,16 @@
     ).to eq(0)
   end
 
+  it "has psql available" do
+    expect(
+      command("psql --version").exit_status
+    ).to eq(0)
+  end
+
+  it "can start postgresql" do
```
corlettb, line 68: This description is a bit misleading as the test doesn't actually check if it can start postgresql. It just checks if the ctl binary/scripts exists.

dark5un: right I will change the description to "has postgresql startup script"

corlettb (approved):
LGTM

